Issue #3, opened 2017-11-13
Add demos for ButtonGroup, button and ripple – one smaple


zdrawku:

rkaraivanov:
closed via #15

PR #15, opened 2017-11-13
Hanastasov/buttons sample

hanastasov:
Issue #3 
Needs to introduce some interactions with the buttons in the sample to make it more appealing

### DIFF
--- a/src/app/buttons/buttons.component.html
+++ b/src/app/buttons/buttons.component.html
@@ -1,0 +1,122 @@
+<div class="sample-wrapper">
+    <section style="flex-flow: row wrap; display: flex; justify-content: center;">
+        <article class="sample-column">
+            <h4 class="sample-title">Flat & Raised Buttons</h4>
+            <div class="buttons-sample">
+                <div class="button-sample">
+                    <button igxButton="flat" igxRipple>Flat</button>
+                </div>
+                <div class="button-sample">
+                    <button igxButton="raised" igxRipple="white">Raised</button>
+                </div>
+                <div class="button-sample">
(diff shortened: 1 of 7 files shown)


src/app/buttons/buttons.component.html
```diff
@@ -0,0 +1,122 @@
...
+                    </button>
+                </div>
+            </div>
+        </article>
+    </section>
+        
+    <section style="flex-flow: row wrap; display: flex; justify-content: center;">
+        <article class="sample-column">
+            <h4 class="sample-title">Exclusive Selection <i class="light">Horizontal</i></h4>
+            <igx-buttongroup [multiSelection]="!multi" [values]="alignOptions">
```
rkaraivanov: This will fail in AOT build. Template variables used should be public.

src/app/buttons/buttons.component.html
```diff
@@ -0,0 +1,122 @@
...
+            </igx-buttongroup>
+  
+            <h4 class="sample-title">Multiple Selection <i class="light">Horizontal</i></h4>
+            <igx-buttongroup [multiSelection]="multi" [values]="fontOptions" [disabled]="false">
```
rkaraivanov: Same as above

src/app/buttons/buttons.component.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+@Component({
+  selector: 'app-buttons',
+  templateUrl: './buttons.component.html',
+  styleUrls: ['./buttons.component.css'],
+  encapsulation: ViewEncapsulation.None
+})
+export class ButtonsComponent implements OnInit {
+  private multi = true;
```
rkaraivanov: Either make them public by default or make getters for each property.

PR #41, opened 2018-01-25
Angular slider sample reworked.

sboykova:
Issue #46


src/app/slider/slider-sample-1/slider-sample-1.component.css
```diff
@@ -0,0 +1,14 @@
+.container {
+    margin-left: 20px;
+    margin-right: 20px;
+    margin-bottom: 30px;
+    margin-top: 30px;
+  }
```
simeonoff: Could we use a shorthand for setting margin properties? Like:
margin: 30px 20px;

src/app/slider/slider-sample-1/slider-sample-1.component.css
```diff
@@ -0,0 +1,14 @@
...
+.lowerInput {
+    clear:both; 
+    float:left;
+}
+.upperInput {
+    clear:none; 
+    float: right;
+}
```
simeonoff: Wrap the inputs in a flex container to avoid using floats for alignment;

src/app/slider/slider-sample-1/slider-sample-1.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="container">
+    <label igxLabel class="titleLabel">Price Range Filter</label>
+</div>
+<div class="container">
+    <igx-slider id="slider" [type]="sliderType.RANGE" [minValue]="0" [maxValue]="1000" [lowerBound]="100" [upperBound]="900" [(ngModel)]="priceRange"></igx-slider>
+    <div class="lowerInput">
+      <label igxLabel for="lowerInput">Lower price:</label>
+      <input igxInput id="lowerInput" type="text" [ngModel]="priceRange.lower | currency" (blur)="updatePriceRange($event.target)" />
+    </div>
```
simeonoff: We have a wrapper class for the input controls. We could use this class to have an input that feels "material". Something like that:
```html
<div class="igx-form-group lowerInput">
        <input igxInput name="lowerInput" type="text" [ngModel]="priceRange.lower | currency" (blur)="updatePriceRange($event.target)">
        <label igxLabel for="lowerInput">Lower price:</label>
 </div>
```

src/app/slider/slider-sample-1/slider-sample-1.component.html
```diff
@@ -0,0 +1,14 @@
...
+    <div class="upperInput">
+      <label igxLabel for="upperInput">Upper price:</label>
+      <input igxInput id="upperInput" type="text" [ngModel]="priceRange.upper | currency" (blur)="updatePriceRange($event.target)" />
+    </div>
+  </div>
```
simeonoff: Same as above.

PR #65, opened 2018-02-15
Rename modules

ddincheva:
Applied all necessary renaming according new version of igniteui-angular.


ddincheva:
All applied changes are already merged in IgniteUI:rename-modules.

PR #72, opened 2018-02-16
Adding igxFor sample with List.

MayaKirova:
Adding sample demonstarting igxFor directive used in an igxList to improve performance.


src/app/igxFor/igxFor.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    }, {
+        avatar: "assets/images/avatar/17.jpg",
+        favorite: false,
+        key: "10",
+        link: "#",
+        phone: "724-742-0979",
+        text: "Ward Riley"
+    }];
+
+  get fo1() {
```
mpavlinov: How about "fo" or "filterOptions" as a name?

src/app/igxFor/igxFor.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+@Component({
+    encapsulation: ViewEncapsulation.None,
+    selector: "igxFor-list",
+    styleUrls: ["./igxFor.component.css"],
+    templateUrl: "./igxFor.component.html"
+})
+export class IgxForComponent implements OnInit {
+
+  public search1: string;
```
mpavlinov: How about "search" as a name?

PR #87, opened 2018-02-21
Fix for theme and fonts

simeonoff:


zdrawku:
We are currently working on the issue with the styles.css file

PR #92, opened 2018-02-22
Updated samples for Linear and Circular Progressbars

ddincheva:
Samples are updated according latest changes in inIgxCircularProgressBarComponent and IgxLinearProgressBarComponent merged with PR [643](https://github.com/IgniteUI/igniteui-angular/pull/643).


zdrawku:
This pull request is going to be merged after we tag 5.2.0, because currently web pack fails to compile due to missing properties like 'IgxTextAlign' etc.

PR #97, opened 2018-02-23
Adding new samples for the TabBar demo page

rmkrmk:
IgniteUI/igniteui-docfx#121


rmkrmk:
The border will be added in the docfx repo.

PR #110, opened 2018-02-26
Update Grid Sample

### LINKED ISSUE #107: Update awesome grid sample
### Description  
We would like for the scroll of the grid to be removed and for the paging to be add again.

A new sample of the Grid will be created which will include virtualization on a later stage.

- [ ] Add paging

ddincheva:
Closes #107


kdinev:
@ddincheva Note my comment in #107

PR #116, opened 2018-02-26
Configure igxFor sample to work using the Stackblitz button

skrustev:


zdrawku:
I am not sure why the push build fails, restarting it right now.

PR #121, opened 2018-02-27
Fixing grid sample alignment and theme

### LINKED ISSUE #112: Fix Grid Sample-header
### Description  
There is no space between slider component and slider label text.

### Result  
image

### Expected result  
What is the expected result after following the steps to reproduce?
image

desig9stein:
close #112


desig9stein:
close #122

zdrawku:
We are going to make some additional style changes, because the Grid is not very responsive now, closing this PR, because it has a lot of changes that we do not mess with.

Close #122

PR #151, opened 2018-04-06
adding category chart overview sample

mharrington622:
I noticed AppModule is kind of big, so I'm interested in feedback on creating Feature Modules for each component. e.g. CategoryChartSamplesModule
Then, continuing to add future samples to the feature module instead of AppModule.  I think routing can be handled per module as well, but I haven't been able to look into it much.  Does anyone know any drawbacks to any of this?


mharrington622:
Also, since the build failed... how should I reference the igniteui-angular-charts package?

gmurray81:
yeah, isn't it not going to scale well to keep adding everything to the main app module @kdinev ? Shouldn't we have different projects for different components maybe? Will be especially important with Stackblitz, I should think, since we want it to have to fetch as little content as possible.

package.json
```diff
@@ -29,6 +29,7 @@
     "gulp": "^3.9.1",
     "hammerjs": "^2.0.8",
     "igniteui-angular": "^5.3.0-beta.1",
+    "igniteui-angular-charts": "^5.3.0-beta.0",
```
gmurray81: I suspect we may need to update this once we've published the final package?

src/app/category-chart/category-chart-samples.module.ts
```diff
@@ -0,0 +1,24 @@
+import { NgModule, ModuleWithProviders  } from '@angular/core';
```
gmurray81: I don't think we should necessarily add more modules structure to the samples, any complexity just makes it more confusing for the user to figure out how they need to extract a sample for the samples browser.

src/app/category-chart/overview/category-chart-overview-sample.component.html
```diff
@@ -0,0 +1,19 @@
...
+		<option>SplineArea</option>
+		<option>StepArea</option>
+		<option>StepLine</option>
+		<option>Waterfall</option>
+	</select>
+</div>
+<igx-category-chart height="500px" width="100%"
+	[dataSource]="data"
+	[chartType]="chartType"
+	isVerticalZoomEnabled="false">
```
gmurray81: I don't think it makes sense to turn this of for catChart. Please remove:

```html
isVerticalZoomEnabled="false"
```

src/app/category-chart/services/data.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from "@angular/core";
+import { Observable } from "rxjs/Observable";
+
+@Injectable()
+export class DataService {
```
gmurray81: Is this degree of complexity worthwhile for a simple sample? Are we trying to leverage a minimal level of best practices or something? The sample is probably easier to follow if it is simply local to the component. @kdinev  will probably have to talk about what their goals have been for the samples. Whether we are optimizing for a minimal level of best practices, or whether we are optimizing for, maybe, being able to view the whole sample in stackblitz without having to navigate between 4 or 5 different files.

src/app/category-chart/overview/category-chart-overview-sample.component.html
```diff
@@ -0,0 +1,19 @@
...
+		<option>Column</option>
+		<option>Point</option>
+		<option>Spline</option>
+		<option>SplineArea</option>
+		<option>StepArea</option>
+		<option>StepLine</option>
+		<option>Waterfall</option>
+	</select>
+</div>
+<igx-category-chart height="500px" width="100%"
```
gmurray81: width 100% is going to look pretty wide in a lot of circumstances possibly, with a static height. What is the recommendation here for if we are hosted in a embedded stackblitz panel? I assume we may want both width and height to be 100%, but in this case you'd need to make sure:

```css
html, body {
   height: 100%;
}
```

src/app/category-chart/overview/category-chart-overview-sample.component.html
mharrington622: Ah, yeah. I meant to remove it, it just annoys me while testing for some reason.

src/app/category-chart/category-chart-samples.module.ts
mharrington622: I initially did this as it's the recommended way to load Feature Modules using services (using ModuleWithProviders and forRoot()).  It prevents multiple instances of the service from being created, since services are globally scoped. If we need to rethink how the whole design will work, that is fine. I think we should wait to hear from Konstantin on the optimization comment below before I go making any more changes.

src/app/category-chart/overview/category-chart-overview-sample.component.html
mharrington622: I see we have `.sample-wrapper { height: 100%; }`, but it looks like it's used in each individual sample.  Which is a bit strange, but I can make that change.  Is this repo supposed to be more of a collection of samples as opposed to being a samples browser?

I would also think we would want the html and body height set on a higher application-level, but I don't currently see any CSS doing that.

package.json
kdinev: Yes, the update is performed every time a new package is published. We do a PR with the new package version and test that everything is OK before the update.

kdinev:
@gmurray81 The app module in the project is independent for what is sent to the stackblitz API. You can verify this with the list's stackblitz button:

https://stackblitz.com/angular/jnlqndjvpae?file=app%2Fapp.module.ts

There shouldn't be different projects for different components.

gmurray81:
@kdinev Should we be indirectly accessing data through services though? Doesn't that make it harder to digest the sample? Does the tooling that extracts just the relevant files for stackblitz like it if there are indirect dependencies?

Is there a good way to test the stackblitz api piece before everything is published?

kdinev:
@gmurray81 Accessing the data through services is the preferred way, because this is what customers would use in their applications. Angular is complex and it's expected that things are located in multiple files. As for the stackblitz testing, there is a way and @zdrawku and his team will walk you through.

tiliev:
@gmurray81, per each sample we create a configuration of what should be included in the sample and how the app module of the sample should look like. You can take a look at two of the most complex samples we have - [the grid samples configs](https://github.com/IgniteUI/igniteui-angular-samples/blob/master/live-editing/configs/GridConfigGenerator.ts). Our StackBlitz engine reads these configs and produces the sample, which is a stand-alone working Angular app by itself.
We do not have a full automation of the generation of these samples because we think it would be an overkill for this task, plus we should take care of all kind of scenarios. The only overhead of our approach is we should add a configuration per each sample we want on StackBlitz.
As far as the testing goes - yes, you could test the sample locally - you just have to set up the [docs repo](https://github.com/IgniteUI/igniteui-docfx), which does the actual calls to the StackBlitz API.
Let me know if you have further questions or propositions for improvement. Thanks.

PR #156, opened 2018-04-11
initial financial chart overview sample

mharrington622:


src/app/financial-chart/services/financial-data.service.ts
```diff
@@ -0,0 +1,21 @@
+import { Injectable } from "@angular/core";
+
+@Injectable()
+export class FinancialDataService {
+
+    public getData(): any[] {
```
gmurray81: I think you may want to have more data here.

mharrington622: I've got 5 years worth of data ready to swap in.  How much are you thinking?

gmurray81: Hmm... not sure. why don't I just approve it in this state, you maybe we need to start a conversation as to how much to throw in there with stakeholders.

gmurray81:
@mharrington622 can you fix the linting error here, so we can start integrating all these samples together.

PR #171, opened 2018-04-16
Add grid samples for columns pinning and remote virtualization

dkamburov:


mpavlinov:
@simeonoff Can you suggest a better "Pin" icon. There is no "Pin" icon in the Material icons.

mpavlinov:
@dkamburov The remote sample data is not enough. There are two few records and every scroll generates a remote request. Try using the data from the "Performance Sample" or part of it.
The Pinning sample has the icons on a new line in Firefox and on Chrome the horizontal scroll is very tiny (maybe a product issue). Also the column headers should be changed, because now are used the column keys.

zdrawku:
Everything looks good, except the stackblitz remote virtualization sample. I cannot see the rendered rows:

image

mpavlinov:
@zdrawku: Yep. There is a Ajax request to get the data from remote, but the StackBlitz is blocking it. Still don't have a solution for it.

zdrawku:
Okay @mpavlinov, since this PR is related for both Pinning and Remove Virtualization samples, I am approving it (positive verify) because everything else is good. 

I leave the desicion to you, whether a separate issue is going to be created and merge this PR, or you are going to add the fix to this PR and merge it later.

mpavlinov:
Related #221

PR #172, opened 2018-04-16
adding category chart generator

mharrington622:


live-editing/SharedAssetsGenerator.ts
```diff
@@ -19,9 +19,11 @@ const SAMPLES_PACKAGE_DEPENDENCIES = [
     "core-js",
     "hammerjs",
     "igniteui-angular",
+    "igniteui-angular-charts",
```
gmurray81: @mharrington622 I thought you were going to figure out how to use a different filter for the charts stuff?

And I don't think we want the non-charts samples to have to load the charts stuff.

mharrington622: Yeah, I just haven't gotten to that yet. Pam was asking about the stackblitz button, so I figured I would at least get this much committed first.  I'm still working on getting only the dependencies that are needed.

gmurray81: I'm not positive that we can leave this in an intermediate state since it will create a bad experience if it ends up being left that way. I think its best to fix it fully first.

tiliev: @gmurray81 & @mharrington622, you could log this as a separate issue int this repository and assign it to me. I can take care of it. I can introduce a new property to the sample configuration specifying which packages to have as dependencies.
Side question - is this package (igniteui-angular-charts) going to stay public? I thought it will be private.

gmurray81:
No, it'll be public, but not free. I know Mike started on this effort and
may have been almost finished? so maybe you want to wait for his pull
request? On a side note, the reason we are bothering to have a separate set
of dependencies is ilwe saw that some of the dependencies of the OSS
control samples take a long time for stackblitz to load, while the
dependencies for the chart load pretty quickly.

One of the other things I was asking Mike to look at was which dependencies
were slowing the load for you guys.

mharrington622:
This should be merged with https://github.com/IgniteUI/igniteui-docfx/pull/290.

As far as loading performance goes, it seems like stackblitz will package resources when the project is saved on their end and will send those packed resources to the browser.  Since we're using the API, we don't actually save the projects, so load times are taking longer.  I am going to be testing if using github imported projects will help.

live-editing/configs/CategoryChartConfigGenerator.ts
```diff
@@ -0,0 +1,31 @@
...
+import { IgxCategoryChartModule } from "igniteui-angular-charts/ES5/igx-category-chart-module";
+import { CategoryChartOverviewComponent } from "../../src/app/category-chart/overview/category-chart-overview-sample.component";
+
+export class CategoryChartConfigGenerator implements IConfigGenerator {
+    generateConfigs(): Config[] {
+        let configs = new Array<Config>();
+        let deps = new Array<string>();
+
+        deps.push("@angular/forms");
+        deps.push("@angular/animations");
```
gmurray81: Don't we not want forms or animations?

live-editing/configs/CategoryChartConfigGenerator.ts
```diff
@@ -0,0 +1,31 @@
...
+        deps.push("igniteui-angular");
```
gmurray81: Or igniteui-angular, for that matter...

gmurray81: not sure what the zip stuff is for either?

PR #177, opened 2018-04-16
added financial-chart samples for performance

HUSSAR-mtrela:
and panes, and titles features

TFS Task# 252213


gmurray81:
what's "performance" about the performance one?

PR #194, opened 2018-04-17
added FC high-volume and update performance

HUSSAR-mtrela:
- added new data services for generating prices
- added FC high-volume sample
- added FC high-frequency sample
- added FC axis-types sample (yAxisMode is not working)
- update FC performance sample


src/app/financial-chart/axis-types/financial-chart-axis-types.component.ts
```diff
@@ -0,0 +1,37 @@
+import { ChangeDetectionStrategy, Component } from "@angular/core";
+import { StockDataService } from "../services/stock-data.service";
+
+import { IgxPercentChangeYAxisComponent
+} from "igniteui-angular-charts/ES5/igx-percent-change-y-axis-component";
+import { IgxPercentChangeYAxisDynamicModule
+} from "igniteui-angular-charts/ES5/igx-percent-change-y-axis-dynamic-module";
+import { IgxPercentChangeYAxisModule
```
gmurray81: You were supposed to import this into the samples module and add it to the imports array...

gmurray81: Why do you even have any of these modules imported here?

src/app/financial-chart/high-frequency/financial-chart-high-frequency.component.ts
```diff
@@ -0,0 +1,177 @@
+import {
```
gmurray81: Having a 10ms update sample for finchart doesn't make any sense. Remove this until this is rethought.

gmurray81: I don't thing having that frequent updates and an for counter makes sense for finchart where it doesn't have that kind of high frequency data, at least under usual conditions. Sample needs more thought.

src/app/financial-chart/high-volume/financial-chart-high-volume.component.html
```diff
@@ -0,0 +1,6 @@
+<igx-financial-chart height="100%" width="100%"
+	[dataSource]="data"
+	chartType="line"
```
gmurray81: Why are you using line here?

gmurray81: Is this the hourly one? Line is ok for that, but the daily should use candle.

gmurray81: Ok this is the hourly one I guess? Ignore me, then.

src/app/financial-chart/services/generate-hourly-prices.service.ts
```diff
@@ -0,0 +1,50 @@
+/* tslint:disable:no-trailing-whitespace max-line-length */
```
gmurray81: Please stop disabling the linting rules unless really necessary.

src/app/financial-chart/services/generate-ohlc-prices.service.ts
```diff
@@ -0,0 +1,60 @@
+/* tslint:disable:no-trailing-whitespace max-line-length */
```
gmurray81: Please stop disabling the linting rules unless really necessary.

src/app/financial-chart/high-volume/financial-chart-high-volume.component.html
HUSSAR-mtrela: yes, that one is using line series because it has data with 1-hour interval which produces a lot of data points.

PR #195, opened 2018-04-17
More generators

mharrington622:


mharrington622:
should be merged with https://github.com/IgniteUI/igniteui-docfx/pull/290

PR #243, opened 2018-04-20
Fix column resizing sample

### LINKED ISSUE #230: Column Resizing  sample issues
### Description  

1. For captions are used field names
2. If I use the keyboard to scroll down the header disappears (also there is no vertical scrollbar). If I scroll up again the header cannot get visible again.

Note: I test the sample in docfx, so 2 can be observed there

image

SAndreeva:
Closes #230


mpavlinov:
@SAndreeva: Please add spaces between words in the headers.

PR #269, opened 2018-04-23
update package to rtm version

gmurray81:


gmurray81:
@kdinev I'm not sure why the lock file is so volatile? Do we all have different npm versions or something?

PR #271, opened 2018-04-24
Adding width to the igxGrid exporters samples.

rmkrmk:
- adding width to the igxGrid exporters samples.
- adding a new line at the end of the file.


src/app/export-csv/csv-export-sample-1/csv-export-sample-1.component.html
```diff
@@ -1,4 +1,4 @@
 <p>
-  <igx-grid #igxGrid1 [data]="localData" [autoGenerate]="true" height="220px"></igx-grid>
+  <igx-grid #igxGrid1 [data]="localData" [autoGenerate]="true" width="100%" height="220px"></igx-grid>
```
kdinev: Isn't the width 100% by default?

rmkrmk: Yes, it looks it is by default. We decided to make changes to the igxGrid's parent container to fix the width issue.

Issue #234, opened 2018-04-20
Some samples are requiring the "immediate" package when running on Stackblitz


rmkrmk:
### Description  
Some samples are requiring the "immediate" package when running on Stackblitz and the site asks for installing it.

### Steps to reproduce  

1. Run docfx samples
2. Open "Services" -> any sample

### Result  
The site asks for installing the "immediate" package.

### Expected result  
The site should not be asking for installing the package.

### Attachments  
![image](https://user-images.githubusercontent.com/11978058/39048163-278ce7a4-44a5-11e8-93ba-a88577a557c2.png)

tiliev:
I asked the guys from StackBlitz regarding this issue (https://github.com/stackblitz/core/issues/424).

tiliev:
`jszip` calls `require('immediate')` in the source without declaring it as a dependency - [link](https://github.com/Stuk/jszip/blob/a4138a23cb9b8eb773c19af7d36d5ce2bcf4bfda/dist/jszip.js#L4565). Closing this as a third-party issue.

PR #288, opened 2018-04-25
Added immediate package to CSV/Excel live editing samples

tiliev:
Fixing CSV/Excel samples until #234 is resolved.


hanastasov:
I get error in the console for samples under Services:
Uncaught TypeError: Cannot read property 'sampleDependencies' of undefined

PR #290, opened 2018-04-25
style(crm-grid): prevent styles leaking

desig9stein:
Prevent styles to leak outside from the grid sample
Related to #249


IvayloG:
Let's wait for this one as some additional changes are required to work in the igniteui-cli app host and rework could be needed.

PR #300, opened 2018-04-27
Fixed the IE11 XSS prevention

### LINKED ISSUE #291: grid.component StackBlitz sample does not load in IE11
### Description  
Clicking on the StackBlitz button for the [DataGrid](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid.html) sample while using IE11 does not load correctly.

### Steps to reproduce  

1. Go to [DataGrid Sample](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid.html) using IE11.
1. Click on **View on StackBlitz** button.

### Result  
The sample on StackBlitz does not load and IE throws an XSS error.

### Expected result  
No errors should show and the sample should run.

### Notes
I have only noticed this issue with this specific sample, so I'm wondering if the XSS error is due to the randomuser.me dataservice.  It doesn't even look like this service is used in the current sample, so we can try removing it to see if that helps? This is related to IgniteUI/igniteui-docfx#271.

tiliev:
"xlink:href" causes the IE 11 to block the request from our docfx application to StackBlitz, even though we have disabled the XSS check with a header. Switching to img tag.
Closes #291.


zdrawku:
I am adding a reminder for this issue, to check it once the changes are present on staging and production

PR #327, opened 2018-05-10
fix(grid-filtering): set percentage width to the last column

### LINKED ISSUE #326: Last columns filtering icon is not visible
Go to https://staging.infragistics.local/products/ignite-ui-angular/angular/components/grid_filtering.html

### Result  
Last columns is not fully visible due to https://github.com/IgniteUI/igniteui-angular/issues/1245

### Expected result  
Last column to be fully visible

We can resolve the issue on application level

Aleksandyr:
Closes #326


mpavlinov:
Please also make PR in branch `5.3.x`

PR #353, opened 2018-05-22
docs(grid): Added samples for remote sorting and filtering #272

### LINKED ISSUE #272: Add Remote Sorting and Filtering to the "Remote Virtualization" sample
### Description  
It's nice to showcase remote Sorting and Filtering for the "Remote Virtualization" sample.

sboykova:
Closes #272

### DIFF
--- a/src/app/grid/grid-remote-filtering-sample/remote-filtering-sample.component.ts
+++ b/src/app/grid/grid-remote-filtering-sample/remote-filtering-sample.component.ts
@@ -1,0 +1,176 @@
+import { HttpClient } from "@angular/common/http";
+import { ChangeDetectorRef, Component, Injectable, OnInit, ViewChild } from "@angular/core";
+import { IgxGridComponent } from "igniteui-angular/grid/grid.component";
+import { IForOfState, STRING_FILTERS } from "igniteui-angular/main";
+import { BehaviorSubject, Observable } from "rxjs";
+
+const EMPTY_STRING: string = "";
+export enum FILTER_OPERATION {
+    CONTAINS = "substringof",
+    STARTS_WITH = "startswith",
+    ENDS_WITH = "endswith",
+    EQUALS = "eq",
+    DOES_NOT_EQUAL = "ne",
+    DOES_NOT_CONTAIN = "not substringof"
+}
+
+@Injectable()
+export class RemoteFilteringService {
+    public remoteData: Observable<any[]>;
+    private _url: string = "https://www.igniteui.com/api/products";
+    private _remoteData: BehaviorSubject<any[]>;
+
+    constructor(private http: HttpClient) {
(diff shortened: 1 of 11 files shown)


mpavlinov:
@sboykova: In #272 I was describing that we add remote sorting and filtering to the current virtualization samples. I'm ok to have 2 different samples, but now the topic seems to load very, very slow because it should load 2 more samples.

mpavlinov:
@sboykova You should remove `angular-in-memory-web-api` from the packages.

mpavlinov:
@sboykova For "Remote Sorting" sample each click on the sorting header results in 2 remote requests. The first one does not have any query params.
image

mpavlinov:
@sboykova For "Remote Filtering" sample each search results in 3 remote requests. 
image

src/app/grid/grid-remote-filtering-sample/remote-filtering-sample.component.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    constructor(private http: HttpClient) {
+        this._remoteData = new BehaviorSubject([]);
+        this.remoteData = this._remoteData.asObservable();
+    }
+
+    public getData(virtualizationArgs?: IForOfState, filteringArgs?: any, cb?: (any) => void): any {
+        return this.http.get(this._url).subscribe((json: any) => {
+            const totalCount = json.length;
+            this.http.get(this.buildDataUrl("", filteringArgs)).subscribe((filteredData: any) => {
```
mpavlinov: @sboykova We should send only one request to the server.

src/app/grid/grid-remote-sorting-sample/remote-sorting-sample.component.ts
```diff
@@ -0,0 +1,136 @@
...
+    private _url: string = "https://www.igniteui.com/api/products";
+    private _remoteData: BehaviorSubject<any[]>;
+
+    constructor(private http: HttpClient) {
+        this._remoteData = new BehaviorSubject([]);
+        this.remoteData = this._remoteData.asObservable();
+    }
+
+    public getData(virtualizationArgs?: IForOfState, sortingArgs?: any, cb?: (any) => void): any {
+        return this.http.get(this._url).subscribe((json: any) => {
```
mpavlinov: @sboykova We should send only one request to the server.

mpavlinov:
@sboykova There is a problem when filter. The remote request returns: "Message:"The query specified in the URI is not valid.""
Steps to repro:
1. Open http://localhost:4200/grid-remote-filtering
2. Search Rating column condition "Equals", search value = 1
3. Observe the network request (there is an error).

Also some of the conditions for numeric columns are not working. For example "Greater than" for "Unit Price".

PR #356, opened 2018-05-23
refactor(card): update people cards with fake data

simeonoff:


ddincheva:
@simeonoff  please update the image url for card-sample-2 and card-sample-3; also when you open card sample vertical scroll is missing so the two of the cards are partly cut  and the latest picture in not visible at all; 
card

hanastasov:
I will make the requested changes since Simo is out of office for the week.

hanastasov:
All changes are PRed with this https://github.com/IgniteUI/igniteui-angular-samples/pull/365

Closing the current PR.

PR #365, opened 2018-05-29
change card samples images and celebrities names 

### LINKED ISSUE #363: Change card sample images
1) Card sample images and names of celebrities should be changed.
2) Also update the image url for card-sample-2 and card-sample-3; also when you open card sample vertical scroll is missing so the two of the cards are partly cut and the last card is not visible at all;
card

hanastasov:
Closes #363

- [x] Change card samples images
- [x] Change card samples celebrities names
- [x] Remove overflow: hidden on body


zdrawku:
I am approving this in order to check it on staging, where we can observe the latest change of overflow configuration that is handled on docfx template level:
https://github.com/IgniteUI/igniteui-docfx-template/issues/4

image

PR #410, opened 2018-06-22
Adding chip sample

alexalex972:


alexalex972:
Currently, the build will fail because IgxChip is not published.

mpavlinov:
@alexalex972 Keyboard navigation in the dropdown doesn't work for me. Can you check it out.

PR #419, opened 2018-06-27
[Merge after beta tag version update] Filter changes and collapse template fix

zdrawku:
This PR should be merfed after we update the igniteui-version to latest


zdrawku:
Duplicate

PR #539, opened 2018-07-17
fix(columnHiding): adjust columnHiding's height for all browsers #538

### LINKED ISSUE #538: 'Display order' radio button is not checked initially.
### Steps to reproduce  

1. Launch Ignite UI for Angular samples
2. Navigate to Data Grid > Column hiding
3. Scroll to the last sample
4. Note the display order radio buttons 

### Result  
None of the radio buttons is checked.

### Expected result  
'Display order' radio button should be checked.

### Attachments  
image

tachojelev:
Closes #538

The display order binding does not work initially since the enum values do not have respective string values. After we modify the **ColumnDisplayOrder** enumeration by assigning these string values, this issue will get automatically fixed.
In addition, in this PR I have adjusted the height of the columnHiding component, so it fits correctly in its parent container, along with the radio buttons beneath it (the ones that are responsible for the display order).


tachojelev:
In order to verify this PR, the radio buttons for the 'display order' in the sample should be positioned beneath the columnHiding component, but inside the columnHidingContainer (a div element) for Chrome, Firefox, IE11 and Edge .

rmkrmk:
I confirm that now the radio buttons for the 'display order' are located beneath the columnHiding component, but inside the columnHidingContainer (a div element) for Chrome, Firefox, IE11 and Edge.

PR #542, opened 2018-07-17
Adding basic paste from excel sample - 6.1.x

### LINKED ISSUE #528: Create a paste from Excel sample
### Description  
We need to demonstrate paste from Excel into the igxGrid

MayaKirova:
Closes #528


mpavlinov:
@MayaKirova If I want to update only one column the data from the other columns is deleted and the cells get empty.

mpavlinov:
@MayaKirova Paste doesn't work on Firefox browser.

Issue #550, opened 2018-07-19
Create UI navigation for the samples' components


tachojelev:
### Description  
Create UI navigation for the samples' components

PR #556, opened 2018-07-24
Modify home page and add new images

tachojelev:
Modify home page and add new images #550


src/app/home/home.component.html
```diff
@@ -1,18 +1,81 @@
...
-    </div>
-</div>
\ No newline at end of file
+<div class="home">
+    <section class="header">
+        <h2 class="title">
+            Welcome to Ignite UI for Angular!
+        </h2>
+        <p class="description">
+            Browse the left menu to see the complete set of samples, including source code, for each control. You’ll see the full array
```
gedinakova: Remove "including source code" as we are not showing the actual source code in the samples.

src/app/home/home.component.html
```diff
@@ -1,18 +1,81 @@
...
+            of capabilities in Ignite UI for Angular, all the different features for the controls, how to implement those
```
gedinakova: "how to implement those features" is not shown either. One should actually open the project in IDE to see that.

src/app/home/home.component.html
```diff
@@ -1,18 +1,81 @@
...
+        </div>
+        <div class="installDescription">
+            <p>
+                If you are already familiar with node.js and using a CLI, you can skip the videos and type:
+            </p>
+            <span class="code-snippet">
+                npm install igniteui-cli –g
+            </span>
+            <p>
+                in Terminal, PowerShell of Bash to install our CLI. We include an alias, so just type:
```
gedinakova: "of Bash" should be "or Bash".

PR #557, opened 2018-07-25
Add setter to igxDropDownItem isSelected property

wnvko:


src/app/overlay/overlay-scroll-2/overlay-scroll-sample-2.component.ts
```diff
@@ -1,22 +1,23 @@
-import { Component, ElementRef, Inject, ViewChild } from "@angular/core";
+import { Component, ElementRef, Inject, OnInit, ViewChild } from "@angular/core";
```
ViktorSlavov: These changes are already present in `master` branch, so I don't think they are needed in the PR.

PR #563, opened 2018-07-25
fix(chip): Fix button not visible

### LINKED ISSUE #562: `Send` button not visible in chip sample
### Description  
Chiparea gets too big when new chip is added and the `Send` button is not visible.

### Steps to reproduce  

1. Open the chip sample https://www.infragistics.com/products/ignite-ui-angular/angular/components/chip.html
2. Add a new email
3. Observe the send button

### Result  
You aren't able to see the `Send` button entirely. If you add more emails the button goes completely hidden.

### Expected result  
The button should be visible during the whole time.

alexalex972:
Closes #562.


PavlovVasil:
Setting the .sample-holder overflow to  "overflow: scroll" makes the scrollbar always visible - even when it's not needed: 

sketch


 I would change it to "overflow-y: auto" instead.

PR #569, opened 2018-07-27
Combo - Remote Data - Change RemoteSample to use Odata V4 for data source - 6.1.x

ViktorSlavov:


Lipata:
@sboykova, @mpavlinov  it seems the [oData ](https://services.odata.org/V4/Northwind/Northwind.svc/Products) service is working well in docfx and with StackBlitz.

PR #608, opened 2018-08-16
Improve the samples for docfx

### LINKED ISSUE #607: Update the IgxTextHighlightDirective samples for docfx
### Description  
The current samples for the IgxTextHighlightDirective do not look very well in the docx. Both samples should move the searchbox on top and also reduce the text on the screen.

DDavidkov:
Closes #607


tachojelev:
It would be nice to have a bigger top margin for the searchboxes and see their top border:
image

tachojelev:
The texthighlight sample components do not seem to have config instances in a ConfigGenerator class, which is necessary for running the samples in stackblitz.
image

PR #657, opened 2018-10-01
Created a drag and drop sample.

MartinKamenov:


mpavlinov:
@MartinKamenov I think it will be better to have the images shuffled in the main grid and the user just need to re-arrange them.

StefanIvanov:
> @MartinKamenov I think it will be better to have the images shuffled in the main grid and the user just need to re-arrange them.

This is very accurate, especially since the grid below is implemented with 9 dragdrop areas and we show dragging from one area to another. However to make such a puzzle work we'd need to leave on piece of the picture out to always provide and area to drop to.

mpavlinov:
@MartinKamenov The functionality looks good to me. The only issue I see are the borders of the tiles. You need to have only 1px border across the puzzle. Not it's not consistent.

mpavlinov:
@MartinKamenov Also, please use a dialog for the "congratulations" message.

mpavlinov:
We also need to add a StackBlitz configuration. I'll add it.

PR #678, opened 2018-10-17
Resize avatars inside the chip sample to not overflow out of the chip.

skrustev:
Related to the refactoring of the chips from  IgniteUI/igniteui-angular/issues/2712


mpavlinov:
@skrustev I don't think it's related with your changes, but I cannot select an e-mail from the suggested e-mail using a keyboard.

PR #695, opened 2018-10-26
excel library samples

HUSSAR-mtrela:
- splitted CC annotation sample based on discussion with Graham
- updated DV packages to v6.2.0-beta.3
- added init samples for excel library


HUSSAR-mtrela:
@pamelabrasil  I will need to re-work this PR

PR #698, opened 2018-10-29
packages update and chart fixes

HUSSAR-mtrela:
- updated DV packages to v6.2.0-beta.3
- corrected FC samples
- spitted CC annotation samples


mharrington622:
If we're going to split out different samples, we should probably create a data service to provide the data across the different samples instead of hardcoding the data in each sample.

tiliev:
@mharrington622, the live editing engine cannot reference files outside of the sample folder. This is currently not supported, that's why we are copy-pasting the data for the time being :cry:

PR #715, opened 2018-11-02
updated packages and set vendorSourceMap

HUSSAR-mtrela:
set vendorSourceMap for Excel Library


zdrawku:
@HUSSAR-mtrela igniteui-angular version was updated and this is causing the conflicts, pull latest and perform npm i, this will resolve the conflicts

PR #716, opened 2018-11-03
SB fixes for importing Excel Library modules

HUSSAR-mtrela:


HUSSAR-mtrela:
@zdrawku we are changing the current configuration because building Excel samples takes longer than default 10min times out.

zdrawku:
@HUSSAR-mtrela I assume travis wait and vendorSourceMap are used for the same? What about the CLI package, it should be the same version as the other dependencies.

Also, I've noticed there is a file that is definind different functions (excel-utility) and it is not used anywhere, is it still "in-development"?

@HUSSAR-mtrela it would be nice if you could add more descriptive note, what exactly is the goal of the changes in the PR, because we need to test ot and ensure it wond break anything.

HUSSAR-mtrela:
@zdrawku I forwarded you an email thread about these changes. The excel-utility is not used yet because first we need to resolve build issues related to using Excel library. Once we do this then most Excel samples will be using this excel-utility.

PR #733, opened 2018-11-06
Setting max height to the IgxCombo in Grid editing sample.

### LINKED ISSUE #730: In Grid editing when you Click add new Product Button you have a Combo in the Dialog which goes outside of the view
### Description  
In Grid editing when you Click add new Product Button you have a Combo in the Dialog which goes outside of the view

### Steps to reproduce  

1. Open Grid Edit sample
2. Click add new Product Button 
3. Click Available @ ComboBox

### Attachments  
image

HristoP96:
Closes #730


HristoP96:
The only resolution for this bug is to set higher level position for the combo in  in the dialog.

The combo-drop-down cannot be modified to render above the combo even if there is no space to render all its content.

This can be logged as a bug since the combo does not provide the ability to render its drop-down  based on the the its container parameters.

zdrawku:
Go ahead and log feature-request for openDropDown direction of the combo.
As for the resolution in this particular case change the position of the dropdown, although keep in mind that the layout should remain user friendly and clean.

PR #787, opened 2018-12-04
feat(TreeGridExport): igniteui-angular#2982 Enabled export via toolbar. 

gedinakova:
https://github.com/IgniteUI/igniteui-angular/issues/2982 
Enabled the export buttons in the toolbar in the main demo only.


zdrawku:
@tachojelev we have pending work here, I guess?

zdrawku:
These are treegrid changes that we are waiting with 7.1.x

PR #806, opened 2018-12-10
Live Editing projects Angular 6 structure migration

### LINKED ISSUE #387: Migrate live editing sample structure to Angular 6
### Description  
The file structure of the live editing samples should be updated to the file structure of Angular 6. Also update the file contents of the new angular.json file.

tiliev:
Closes #387.

Please test it on IE 11 as well.


tiliev:
@hanastasov, this appears to be a new issue and it reproduces only on local. It is not connected to the current PR. I logged it [here](https://github.com/IgniteUI/igniteui-docfx-template/issues/62).

PR #809, opened 2018-12-11
Update igniteui-angular package to version 7.1.0-beta.1

bkulov:
Update igniteui-angular package to version 7.1.0-beta.1


IvayloG:
Currently the “Boston Marathon 2018” Grid sample is not OK as it is not rendered correctly and throws 
_ERROR TypeError: Cannot read property 'length' of undefined._ 
image

PR #816, opened 2018-12-13
Update CRM grid sample to use IgxGridToolbar

rmkrmk:
This PR removes the toolbar created in the sample and uses the toolbar provided in the IgxGrid.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -100,7 +100,7 @@
       "dependencies": {
         "source-map": {
           "version": "0.5.6",
-          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.6.tgz",
+          "resolved": "http://registry.npmjs.org/source-map/-/source-map-0.5.6.tgz",
           "integrity": "sha1-dc449SvwczxafwwRjYEzSiu19BI=",
           "dev": true
         },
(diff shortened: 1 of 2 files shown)


desig9stein:
@gedinakova this is the correct contrast color based on the W3C Accessibility standards. you can check it [here](https://webaim.org/resources/contrastchecker/), instead of changing the text I should probably change the logo color.

gedinakova:
@desig9stein It would look different compared to how the Propeller grid used to look. Not sure what should be done here.

src/app/grid/grid-crm/grid-crm.component.html
```diff
@@ -1,206 +1,166 @@
...
-                                    <li class="dd-list__item" *ngFor="let col of cols">
-                                        <igx-checkbox (change)="toggleVisibility(col)" [checked]="col.hidden">{{col.header}}</igx-checkbox>
-                                    </li>
-                                </ul>
+    <igx-grid #grid1 id="grid1" [data]="localData" [height]="'430px'" [width]="'1180px'" [rowSelectable]="true"
+        displayDensity="cosy" rowHeight="50" [allowFiltering]="true"
+        [showToolbar]="true" [columnHiding]="true" columnHidingTitle="Column Hiding"
+        [columnPinning]="true" [exportExcel]="true" [exportCsv]="true" exportExcelText="Export to Excel" exportCsvText="Export to CSV">
+
+        <ng-template igxToolbarCustomContent let-igxGrid="grid">
```
gedinakova: Remove "let-igxGrid" from the template - it's not necessary.

rmkrmk: Done.

PR #856, opened 2019-01-07
Adding schemas usage to the theme chooser sample.

### LINKED ISSUE #855: Add schemas to theme chooser sample.
Because of this [issue](https://github.com/IgniteUI/igniteui-docfx/issues/876) we should provide a sample with schemas usage

HristoP96:
Closes #855


src/app/theming/styles/themes.scss
```diff
@@ -1,5 +1,6 @@
 // Import the theme utilities first
 @import '~igniteui-angular/lib/core/styles/themes/utilities';
+@import '~igniteui-angular/lib/core/styles/themes/schemas/_index';
```
simeonoff: No need to import this as it's already imported via utilities.

HristoP96: The redundant import is removed

src/app/theming/styles/schemas.scss
```diff
@@ -0,0 +1,188 @@
+@import '~igniteui-angular/lib/core/styles/themes/utilities';
+
+$black-color: #000000;
+$aqua-color: #00BFFF;
+
+$black-theme-palette: igx-palette($primary: $black-color, $secondary: $aqua-color);
```
simeonoff: Ensure you set the $grays property to #fff for dark palettes.

HristoP96: $grays added to the palette

Issue #776, opened 2018-11-30
Add lazy loading to all samples routes


zdrawku:
### Description  
Curently only Excel library samples have `lazy loading`.
[Lazy loading](https://github.com/IgniteUI/igniteui-angular-samples/blob/9ed0f37594e03829b4cf2683caad19dddaa32751/src/app/app-routing.module.ts#L345) should be implemented for all samples

## Bundles
- [x] **Grid CRM** bundle - in progress Hristo
- [x] **Charts** bundle - (https://github.com/IgniteUI/igniteui-angular-samples/pull/868)
- [x] **Gauges** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/861)
- [x] **Excel library** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/869)
- [x] **IgxGrid** bundle - in progress Hristo
- [x] **IgxTreeGrid** bundle  -(https://github.com/IgniteUI/igniteui-angular-samples/pull/877)
- [x] **Lists** bundle - List View + Combo(https://github.com/IgniteUI/igniteui-angular-samples/pull/867)
- [x] **Styling and themes** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/866)
- [x] **Data Entries** bundle - (https://github.com/IgniteUI/igniteui-angular-samples/pull/878)
- [x] **Data Display** bundle - (https://github.com/IgniteUI/igniteui-angular-samples/pull/878)
- [x] **Interactions** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/859)
- [x] **Menus** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/865)
- [x] **Layouts** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/874) 
- [x] **Notifications** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/857)
- [x] **Scheduling** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/863)
- [x] **Services** bundle (https://github.com/IgniteUI/igniteui-angular-samples/pull/875)

tiliev:
We should separate the sample into different modules. The lazy loading will create a `.js` file per each module. These `.js` files will be loaded only if a sample in the module is accessed.
The following bundles could be created:
[...]

Issue #798, opened 2018-12-07
Pie Chart samples do not work in StackBlitz


tiliev:
### Description  
Pie Chart samples do not work in StackBlitz.

### Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular/angular/components/piechart.html
2. Hit any `View on StackBlitz` button.

### Result  
Nothing happens.

### Expected result  
The sample should be opened in StackBlitz.

### Attachments  
![pie-chart](https://user-images.githubusercontent.com/6439482/49658430-b0785480-fa4a-11e8-8a1f-36d736fd7915.PNG)

Issue #738, opened 2018-11-07
No ConfigGenerators for the DoughnutChart samples.


HristoP96:
### Description  
The Doughnut Chart samples are not exported in StackBlitz.
### Steps to reproduce  

Open Infragistics Doughnut Chart section.
[https://staging.infragistics.local/products/ignite-ui-angular/angular/components/doughnutchart.html](https://staging.infragistics.local/products/ignite-ui-angular/angular/components/doughnutchart.html) 

Click on the "View on Stackblitz" button below every Doughnut chart sample.
### Result  
No StackBlitz samples the Doughnut charts.

### Expected result  
The Doughnut charts should have  StackBlitz samples,

PR #868, opened 2019-01-11
Charts lazy loading 776

hanastasov:
Added charts bundle #776


tiliev:
@nrobakova and @hanastasov , please keep in mind we have issues for these - https://github.com/IgniteUI/igniteui-angular-samples/issues/798 and https://github.com/IgniteUI/igniteui-angular-samples/issues/738. It is kind of strange we have StackBlitz buttons without configurations. We either should add configurations or remove the StackBlitz buttons and leave the issues for near future. @zdrawku could decide what we should do.

PR #885, opened 2019-01-16
Set width to searchBox

### LINKED ISSUE #884: Search box is cut on the right when loading the samples home screen
### Description  
Search box is cut on the right when loading the samples home screen.

### Steps to reproduce  

Run the igniteui-angular-samples.

### Result  
Search box is cut on the right.

### Expected result  
Search box is **not** cut on the right.

### Attachments  

image

tachojelev:
Closes #884


tachojelev:
After further investigation, the issue seems to be originating from the **IgxNavDrawer** component and it is reproduced only on **Chrome**.
The **IgxNavDrawerItem** styles seem to be resolved too early in **Chrome**, which leads to items not respecting the vertical scrollbar width and they go under it.

We will fix the current samples issue by replacing setting **max-width** of the search box for the time being.

PR #896, opened 2019-01-22
Fixing Drag and Drop directive sample

PavlovVasil:
Solving the puzzle should not throw an error.


src/app/interactions/drag-drop/drag-drop.component.ts
```diff
@@ -55,41 +55,35 @@ export class DragAndDropSampleComponent implements OnInit {
...
-        const dragIdString = ev.drag.element.nativeElement.id;
-        const dropId = dropIdString.match(/\d+/g).map(Number)[0];
-        const dragId = dragIdString.match(/\d+/g).map(Number)[0];
-
-        const foundIndex = this.findIndexOfDrag(dragId);
-        const firstPartRow = foundIndex.row;
-        const firstPartColl = foundIndex.coll;
+        const dropId = parseInt((ev.owner.element.nativeElement.id).slice(-1), 10);
+        const dragId = parseInt((ev.drag.element.nativeElement.id).slice(-1), 10);
+        const firstPartRow = this.findIndexOfDrag(dragId).row;
```
MartinKamenov: @PavlovVasil 
Using this.findIndexOfDrag twice with same parameter is unnecessary and slower than saving the result from the first call.
I would suggest to stay as it was.
(
const foundIndex = this.findIndexOfDrag(dragId);
const firstPartRow = foundIndex.row;
const firstPartColl = foundIndex.coll;
)

PavlovVasil: I have brought the variable back and fixed the lint error as well.

PR #914, opened 2019-01-29
Updating the fixing the Chip sample

### LINKED ISSUE #822: Chips sample throws "igxOverlay.hide was called with wrong id: undefined" warning
### Description  
Describe the issue.

### Steps to reproduce  

1. Open: https://www.infragistics.com/products/ignite-ui-angular/angular/components/chip.html
2. Drag and drop a chip from the sample
3. Observe the console

### Result  
There is a warning in the console: igxOverlay.hide was called with wrong id: undefined

### Expected result  
There should be no warning in the console

PavlovVasil:
Closes #822 
Please also test the following:

1. Adding emails:
    *  Click on the input, type in some new email, press "Enter". A new Chip should be added.
    *  Click on the input again and try to type in another new email. There was a problem with the dropdown overlay not being removed from the DOM in some cases, and it was covering the input, preventing the user from typing the second time. You should be able to add as many new emails as you want to.

2.  Validation:
    * Click on the input, then click outside of it - the validation should not be triggered and the input should not turn red.


MartinKamenov:
@PavlovVasil The overlay elements stays when dropdown of mail input is shown. This blocks user from writing in the text form and clicking the send button.
- Steps to reproduce.
1. Select the mail input for writing.
2. Press down arrow key twice fast.
3. Try to click on the send button.

* Note
- This is reproduced before this pr.

MartinKamenov:
@PavlovVasil 
Issue was resolved after latest changes.
I would only recommend changing the clickedOutside event name to clickedOutsideDropDown.
This would make it more meaningful.

PR #918, opened 2019-01-30
Prefixing css styles for every sample

### LINKED ISSUE #626: Grids with width in % in IE are broken in StackBlitz
Open any of the grid samples with width in percent in StackBlitz. It is broken.
More information in this issue - https://github.com/IgniteUI/igniteui-angular/issues/1330.

HristoP96:
Closes #626

With these changes, when a sample is opened in StackBlitz(**in IE** ), the global and component styles will be prefixed in order to render the sample properly


live-editing/services/SassCompiler.ts
```diff
@@ -27,7 +29,8 @@ export class SassCompiler {
             includePaths: NODE_SASS_PATHS
         });
 
-        return renderedSassResult.css.toString();
+        return postcss([autoprefixer({browsers: ["last 5 versions", "> 3%"], grid: true})]).
```
tiliev: return should be the last statement of the function

Issue #890, opened 2019-01-18
Add new demo samples


hanastasov:
Think of what can be added to existing samples and what new samples to be added to fulfill the below requirements for new samples:

- [x] Grid properties Save / Load. - [Working sample](https://grid-save-state.stackblitz.io/grid-state) - Demonstrate how a developer can save the grid state in terms of column moving, sorting, filt, etc. when user leaves page or close browser, and have the ability to re-load the state when grid reloads.

- [ ] Responsive UI - Demonstrating various RWD examples, like re-template a grid column, hide column, etc, it will alleviate many q’s, benefiting from media queries too.  Similar to https://www.igniteui.com/grid/responsive-web-design-mode

- [x]  [Working sample](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/paging.html#remote-data) Pager UX - We can modify it further to change button styles, use hyperlinks for page numbers vs. the current UX.  Demonstrating more examples of Pager styling it will help answer q’s. 

- [x] Custom keyboard control - Customizing what happens when Edit Complete (like when Enter key is hit, or Tab key is hit).  For example, going to next cell, cell below, etc.  Demonstrating what API to use to customize key behavior. Done 
https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/keyboard_navigation.html#custom-keyboard-navigation

Additional samples updated to latest version:
[People app](https://infragistics-indigo-design-sample-people.stackblitz.io/)
[Budget app](https://budget-application.stackblitz.io/)

zdrawku:
@StefanIvanov we've finished with almost all examples, the only one left is Responsive UI. Can you guys help us with it. These samples were mentioned by Jason a while ago, that would be nice to have them on the Angular documentation site.

> Responsive UI - Demonstrating various RWD examples, like re-template a grid column, hide column, etc, it will alleviate many q’s, benefiting from media queries too. Similar to https://www.igniteui.com/grid/responsive-web-design-mode

zdrawku:
We are currently working on other application scenarios, closing this one.

PR #933, opened 2019-02-05
Sample demonstrating how to save and load the igxGrid state

hanastasov:
Part of #890


nrobakova:
I have tested the sample and there are some issues:
1. When for a column select 2 filters with operand AND, and reload the page the operand is changed to OR
2. When for a column select 2 filters with operand OR, only the first filter is saved after reload
3. When for column Register Date filter with date, after reload this date is not in correct format
4. When selection is checked only row selection is saved, but cell selection is not.

zdrawku:
@hanastasov, I found the following issues below:
- filtering type is "string" for all columns
- after clearing the localStorage, the state is persisted again, while navigation to About page and back, or by refreshing the page with the refresh button

hanastasov:
> @hanastasov, I found the following issues below:
> 
> * filtering type is "string" for all columns --> fixed with latest push
> * after clearing the localStorage, the state is persisted again, while navigation to About page and back, or by refreshing the page with the refresh button --> state is being saved at the moment you navigate away. This is why afetr clearing the storage, use the reload button to reload the page with clear state.

nrobakova:
@hanastasov can you look at this issues:
1. Can you change the checkbox Selection to Row Selection, because it may be clear that only the row selection is saved.
2. When I select a row, filter with no values, leave page and return to it, the row selector header is checked, but it should not be.

hanastasov:
Please test the Stackblitz config.

PR #938, opened 2019-02-11
Updating igniteui-angular version to 7.2.0-beta.3

mpavlinov:


mpavlinov:
@bkulov The build is failing. It seems that there is a circular dependency that we need to resolve first and mark a new beta.

kdinev:
@mpavlinov Retarget this to a branch different from `master`.

PR #972, opened 2019-02-18
Increased the RAM limit to 8 GB

### LINKED ISSUE #772: Rebuilding with ng 7 throws OutOfMemory exception spontaneously
### Description  
[Related PR](https://github.com/IgniteUI/igniteui-angular-samples/pull/780)

`npm run start` throws our of memory exception

Steps to reproduce:

- Get latest master
- Execute npm run start
- Update a sample, do this 2-3 times
-OutOfMemory exception is thrown

tiliev:
Closes #772.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,10 +2,5 @@
 	"editor.tabSize": 4,
 	"editor.insertSpaces": true,
     "editor.detectIndentation": false,
-    "jscs.enable": false,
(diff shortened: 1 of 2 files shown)


tiliev:
I am also removing the customized VS color scheme. I guess it was committed by mistake by @HUSSAR-mtrela. Actually it looks very good but we should leave the default color scheme.

PR #978, opened 2019-02-19
Updating igniteui-angular version to 7.2.0-beta.3

mpavlinov:


tiliev:
According to the docs `enableProdMode`  will 
```
Disable Angular's development mode, which turns off assertions and other
checks within the framework.
```

It does not make any sense for us to call `enableProdMode` when we are not in `prod` mode.
I created a new project with the latest `@angular/cli` and in their `main.ts` I can see:
```
if (environment.production) {
  enableProdMode();
}
```
which makes sense. I think we can proceed with this approach.

PR #981, opened 2019-02-21
Adding Hierarchical Grid Load on Demand sample.

skrustev:


skrustev:
> @skrustev
> 
> * [x]  The OrderDate format is "1997-08-25T00:00:00Z". Can you change it to En-Us format "mm/dd/yyyy".
> * [ ]  There is something wrong with the horizontal scroll of the Orders child grid. When I expand initially there is a horizontal scroll, which disappears when I expand an order.
> * [x]  Remove `CustomerID`, `OrderID` and `ProductID` columns from the grids.

The second might be an issue in the Hierarchical Grid itself. I will investigate it further.
I will close this PR and commit the sample in the pr #944

PR #990, opened 2019-02-26
Update overlay samples with attach function

wnvko:
All overlay samples will now use `attach` function to obtain overlay ID. Then in show `method` provided ID will be used to show the content.


src/app/interactions/overlay/overlay-main-1/overlay-main-sample-1.component.ts
```diff
@@ -10,13 +8,21 @@ import { CardSample1Component } from "../../../layouts/card/card-sample-1/card-s
...
 export class OverlaySampleMain1Component {
+    private _overlayId: string;
 
     constructor(
-        @Inject(IgxOverlayService) private overlayService: IgxOverlayService
+        @Inject(IgxOverlayService) public overlayService: IgxOverlayService
     ) {
+        //  overlay service deletes the id when onClosed is called. We should clear our id
+        //  also in same event
+        this.overlayService.onClosed.subscribe(() => delete this._overlayId);
```
ViktorSlavov: I think it would be better if the subscription checks for the emitted ID and `delete`s in only if it matches the component's overlayId - it would be better suited for a real application scenario.
The subscription should also be cleared when the component's ngDestroy is called.

wnvko: Thank you for nice point. It should be fixed now in all relevant samples.

PR #1008, opened 2019-02-28
Add web.config for SVG assets

### LINKED ISSUE #781: SVG Icon Samples are not loaded in StackBlitz
### Description  
SVG Icon Samples are not loaded in StackBlitz

### Steps to reproduce  

1. Go to https://www.infragistics.com/products/ignite-ui-angular/angular/components/icon.html
2. Scroll to the bottom of the page where is the SVG Icons samples
3. Click on "View on StackBlitz" button.

### Result  
The icons are not loaded.
There is an error in the console:

"ERROR
Error: Could not fetch SVG from url: https://www.infragistics.com/angular-demos/assets/images/svg/does_not_equal.svg; error: 
Http failure response for (unknown url): 0 Unknown Error"

### Expected result  
The icons are loaded.
No error.

image

bkulov:
Closes #781

Allow SVG files to be accessed outside from Infragistics domain.


kdinev:
@bkulov shouldn't only `*.stackblitz.io` be allowed instead of `*`?

bkulov:
@kdinev this is something @damyanpetev wrote on my machine. I remember we've tested it with `*.stackblitz.io` but it didn't worked for some reason. Maybe @damyanpetev could say more about that.

damyanpetev:
@kdinev @bkulov
Wildcard domains (like *.stackblitz.io) are sadly [not valid](https://stackoverflow.com/a/14006226) values for `Access-Control-Allow-Origin`. It's either a fixed list of origins or only `"*"`.
Could be done with server logic, but the config-only route seems [a bit extreme](https://www.carlosag.net/articles/enable-cors-access-control-allow-origin.cshtml) for 5 or so svg-s.

damyanpetev:
Verified with the issue's Stackblitz sample with links pointing to local IIS 10.0.17763.1

PR #1019, opened 2019-03-01
Fixing scroll position igxGrid Paste From Excel sample

### LINKED ISSUE #883: In sample Paste from Excel when paste data as new record it is not decorated in Italic
### Description  
In sample Paste from Excel when paste data as new record it is not decorated in Italic

### Steps to reproduce  

1. Open https://www.infragistics.com//angularsite/components/grid_paste_excel.html#paste-from-excel-demo
2. Paste data as a new record
3. Scroll down to see the pasted data

### Result  
Data is not decorated in Italic as it is written in the comment

### Expected result  
Data should be decorated in Italic

### Attachments  

kacheshmarova:
Closes #883 
In sample Paste from Excel when paste data as new record it is not decorated in Italic


MartinKamenov:
@kacheshmarova 
When the grid is scrolled to bottom new rows are not marked in italic.

pasteexcel

VMihalkov:
Now when you copy the new rows, they initially appear in italic, but after you scroll up and down, they are not italic anymore

mpavlinov:
@kacheshmarova Can you add "Save Changes", "Undo", "Redo" buttons and use the Updating API.

PR #1022, opened 2019-03-02
Updating to igniteui-angular 7.2.0-rc.1

mpavlinov:


mpavlinov:
@MayaKirova @wnvko There is an error that is caused by a recent change of `outletDirective` being only getter now. This is the commit that changes it https://github.com/IgniteUI/igniteui-angular/commit/b209c8ddfd9314386a93b12b27ee5e28c685745d. Should we also create a setting in the product?

MayaKirova:
@mpavlinov @wnvko Since the original option was not documented (marked as hidden) and seemed to be meant to just retrieve the default outlet in the grid template it didn’t seem necessary to expose a setter.
That said, if the idea behind it is to allow setting a different outlet that can be reused between components  (for example if you have many components that display overlays on the same page) then we should expose a setter and document it.
Let me know what you think.

mpavlinov:
@zdrawku @HristoP96 What do you think?

HristoP96:
@mpavlinov The line causing this error is redundant. 

I will remove it so that the build could be successful.

HristoP96:
The redundant line is removed in this PR - [https://github.com/IgniteUI/igniteui-angular-samples/pull/1025](https://github.com/IgniteUI/igniteui-angular-samples/pull/1025)

PR #1029, opened 2019-03-05
fix(dropdown): update samples to use igxToggleAction, cleanup CSS

### LINKED ISSUE #1028: IgxDropDown - Samples should use igxToggleAction where possible
### Description  
The samples currently use `dropdown.toggle()` method, passing custom overlay settings. The best and easiest way to use the `igx-drop-down` is through the use of `igxToggleAction` directive. Having to pass custom overlay settings through the `toggle` method is clunky and error prone. When making use of 'closeOnOutsideClick, `excludePositionTarget: true` must be passed in order to avoid unexpected behavior. 

 - Remove click events
 - Bind drop down toggles using `igxToggleAction`
 - Clear up .scss and .ts files

This will make the samples (especially the template.htmls) more easily understandable.

### Result  
Drop Down samples are currently using `dropdown.toggle()` method and passing custom overlay settings.

### Expected result  
Drop Down samples should use `igxToggleAction` directive.

ViktorSlavov:
Closes #1028


src/app/data-entries/dropdown/dropdown-menu/dropdown-menu.component.ts
```diff
@@ -17,17 +10,8 @@ import {
     templateUrl: "./dropdown-menu.component.html"
 })
 export class DropdownMenuComponent implements OnInit {
-    @ViewChild(IgxOverlayOutletDirective) public igxOverlayOutlet: IgxOverlayOutletDirective;
-    @ViewChild(IgxDropDownComponent) public menu: IgxDropDownComponent;
-
     public items: any[] = [];
-    public text;
-
-    private positionSettings = {
```
wnvko: Default settings in current sample are start point Right-Bottom and direction Left-Bottom. In updated sample we have Left-Bottom start point and Right-Bottom direction. Should we have same settings in old and updated sample?

wnvko: We had here NoOp scroll strategy and in the new sample it will be Absolute scroll strategy!

src/app/data-entries/dropdown/dropdown-sample-1/dropdown-sample-1.component.ts
```diff
@@ -1,42 +1,14 @@
...
-
-    private _positionSettings = {
-        horizontalStartPoint: HorizontalAlignment.Left,
-        verticalStartPoint: VerticalAlignment.Bottom
-    };
-    private _overlaySettings = {
-      closeOnOutsideClick: true,
-      modal: false,
-      positionStrategy: new ConnectedPositioningStrategy(this._positionSettings),
-      scrollStrategy: new CloseScrollStrategy()
```
wnvko: We had here Close scroll strategy and in the new sample it will be Absolute scroll strategy!

src/app/data-entries/dropdown/dropdown-sample-2/dropdown-sample-2.component.ts
```diff
@@ -1,47 +1,15 @@
...
-
-    private _positionSettings = {
-        horizontalStartPoint: HorizontalAlignment.Left,
-        verticalStartPoint: VerticalAlignment.Bottom
-    };
-    private _overlaySettings = {
-      closeOnOutsideClick: true,
-      modal: false,
-      positionStrategy: new ConnectedPositioningStrategy(this._positionSettings),
-      scrollStrategy: new CloseScrollStrategy()
```
wnvko: We had here Close scroll strategy and in the new sample it will be Absolute scroll strategy!

src/app/data-entries/dropdown/dropdown-sample-3/dropdown-sample-3.component.ts
```diff
@@ -24,20 +16,4 @@ export class DropDownSample3Component {
...
-
-    private _positionSettings = {
-        horizontalStartPoint: HorizontalAlignment.Left,
-        verticalStartPoint: VerticalAlignment.Bottom
-    };
-    private _overlaySettings = {
-      closeOnOutsideClick: true,
-      modal: false,
-      positionStrategy: new ConnectedPositioningStrategy(this._positionSettings),
-      scrollStrategy: new CloseScrollStrategy()
```
wnvko: We had here Close scroll strategy and in the new sample it will be Absolute scroll strategy!

src/app/data-entries/dropdown/dropdown-sample-4/dropdown-sample-4.component.ts
```diff
@@ -1,36 +1,24 @@
...
-    public onSelection(eventArgs: ISelectionEventArgs) {
-        this.value = eventArgs.newSelection.value;
-    }
+    public items: Array<{ field: string }> = [
+        { field: "Option 1" },
+        { field: "Option 2" },
+        { field: "Option 3" }
+    ];
 
     public openDropDown() {
```
wnvko: this will not work now. You cannot change the target position strategy when you are using IgxToggleAction. Also there is no need to check if drop down is collapsed - just call Open - IgxToggleDirective will check everything and will open the drop down only when necessary.

src/app/data-entries/dropdown/dropdown-sample-5/dropdown-sample-5.component.ts
```diff
@@ -58,20 +50,4 @@ export class DropDownSample5Component {
...
-
-    private _positionSettings = {
-        horizontalStartPoint: HorizontalAlignment.Left,
-        verticalStartPoint: VerticalAlignment.Bottom
-    };
-    private _overlaySettings = {
-        closeOnOutsideClick: true,
-        modal: false,
-        positionStrategy: new ConnectedPositioningStrategy(this._positionSettings),
-        scrollStrategy: new CloseScrollStrategy()
```
wnvko: We had here Close scroll strategy and in the new sample it will be Absolute scroll strategy!

src/app/data-entries/dropdown/dropdown-sample-4/dropdown-sample-4.component.ts
ViktorSlavov: This is a call to the drop down's open method, which positions the content directly through the overlay service, bypassing the `igxToggleAction` directive. It will display the drop down properly, though the method itself is not really elegant - as we can't access the `igxToggleAction`'s overlay settings

src/app/data-entries/dropdown/dropdown-sample-2/dropdown-sample-2.component.ts
ViktorSlavov: Thanks for pointing this out! I don't think the scroll strategy is really relevant here though, no need to clutter up the example with custom settings.

src/app/data-entries/dropdown/dropdown-sample-3/dropdown-sample-3.component.ts
ViktorSlavov: Thanks for pointing this out! I don't think the scroll strategy is really relevant here though, no need to clutter up the example with custom settings.

src/app/data-entries/dropdown/dropdown-sample-5/dropdown-sample-5.component.ts
ViktorSlavov: Thanks for pointing this out! I don't think the scroll strategy is really relevant here though, no need to clutter up the example with custom settings.

src/app/data-entries/dropdown/dropdown-menu/dropdown-menu.component.ts
ViktorSlavov: Thanks for pointing that out! I'll pass modified overlay settings through the markup to illustrate how that is done. For the other samples, I think the scroll strategy is not really relevant and passing settings in each of the samples will result in clutter.

src/app/data-entries/dropdown/dropdown-sample-1/dropdown-sample-1.component.ts
ViktorSlavov: Thanks for pointing this out! I don't think the scroll strategy is really relevant here though, no need to clutter up the example with custom settings.

PR #1044, opened 2019-03-07
Updating to igniteui-angular 7.2.0-rc.4

bazal4o:


zdrawku:
After further investigation with the team, it looks like this error is related to the --aot build and label input

ERROR in : Can't bind to 'label' since it isn't a known property of 'igx-select-item-group'.

PR #1054, opened 2019-03-11
Replace drop down with select in cascading scenario sample

### LINKED ISSUE #1041: Replace drop down with select in cascading scenario sample
### Description  
Replace both drop-down components with select in Combo -> Cascading Scenario sample.

Lipata:
Closes #1041


ViktorSlavov:
The PR is missing an update to `ComboConfigGenerators.ts`, so the cascarding sample StackBlitz won't work. Other than that, everything looks great!

ViktorSlavov:
The StackBlitz sample works fine with the changes to the config file. The file structure is a bit off, but I can't find any fault in the configuration:
cascarding-stackblitz

PR #1084, opened 2019-04-04
Mtp igniteui angular samples spreadsheet

igMPeterson:
Spreadsheet sample shells and routing setup


src/app/spreadsheet/ExcelFileUtilities.ts
```diff
@@ -0,0 +1,81 @@
+import { saveAs } from "file-saver";
```
HUSSAR-mtrela: this file already exists in utilities so re-use it instead of creating duplicate

HUSSAR-mtrela:
@pamelabrasil @zdrawku 
this PR looks good to me and I think you can merge it.

PR #1104, opened 2019-04-11
Replacing all the images for HGrid samples with new ones

PavlovVasil:
Closes: #1112


mpavlinov:
@PavlovVasil Please add 4 more images so that they don't duplicate.

PavlovVasil:
@mpavlinov Please test if it is OK now.

PR #1149, opened 2019-05-10
Multi-Row-Configurations sample

MartinKamenov:


mpavlinov:
@skrustev The sample needs work. For example there is no padding between the buttons. There are a lot of white spaces.
I suggest that we have a pre-configured layout that's assembly the other sample for MRL.

PR #1150, opened 2019-05-10
time picker: add dropdown retemplate sample

SAndreeva:


src/app/scheduling/timepicker/timepicker-sample-6/timepicker-sample-6.component.ts
```diff
@@ -0,0 +1,19 @@
...
+})
+export class TimePickerSample6Component {
+
+    public today: Date = new Date(Date.now());
+
+    public onBlur(inputValue, timePickerValue) {
+        const parts = inputValue.split(":");
+
+        if (parts.length === 2) {
+            timePickerValue.setHours(parts[0], parts[1]);
```
bkulov: You can add some validation if the user's input is correct. If the the user's input is incorrect there will be an exception and we don't want that in our samples.

sboykova:
Drop-down appears at the right when the sample is run in the samples browser - docfx.

PR #1159, opened 2019-05-14
Add new row drag sample

### LINKED ISSUE #1158: Additional sample for grid row drag
Use row dragging in an application scenario

Lipata:
Closes #1158


ViktorSlavov:
I couldn't get the StackBlitz button to work for me, but the config seems fine.
Everything else works great!

PR #1160, opened 2019-05-14
Add sample with DisplayDensity for the IgxButtonGroup

mtsvyatkova:
IgniteUI/igniteui-angular#4532


src/app/data-entries/buttonGroup/button-group-sample-5/button-group-sample-5.component.html
```diff
@@ -0,0 +1,5 @@
+<article class="sample-column">
+    <igx-buttongroup [multiSelection]="false" [values]=displayDensities (onSelect)="selectDensity($event)"
```
gedinakova: No quotations around displayDensities.

mtsvyatkova: Just added them. Thanks

src/app/data-entries/buttonGroup/button-group-sample-5/button-group-sample-5.component.ts
```diff
@@ -0,0 +1,27 @@
...
+export class ButtonGroupSample5Component implements OnInit {
+
+    public rippleColor = "grey";
+
+    public displayDensity = "comfortable";
+    public displayDensities;
+
+    public ngOnInit() {
+        this.displayDensities = [
+            { label: "cosy", selected: this.displayDensity === "cosy", togglable: true },
```
tachojelev: We should define the objects in ascending/descending order according to the density, so they get displayed in that order in the button-group as well. For example: "**compact**"->"**cosy**"->"**comfortable**"

PR #1165, opened 2019-05-17
Set display density and use ExcelStyleFiltering on the finjs demos

### LINKED ISSUE #1164: Use Excel Style Filtering on the finjs demos

hanastasov:
Closes #1164


hanastasov:
Needs to be tested with 7.3.1, once tagged.

Issue #782, opened 2018-12-03
Stackblitz configuration: The dependencies for this project couldn't be resolved


lalo-mx:
### Description  
Could not open samples on StackBlitz

### Steps to reproduce  

1. Open StackBlitz import page for samples https://stackblitz.com/github/IgniteUI/igniteui-angular-samples/.
2. An error is shown: The dependencies for this project couldn't be resolved.

### Result  
An error: The dependencies for this project couldn't be resolved.

### Expected result  
The samples run and be displayed.

### Attachments  
![image](https://user-images.githubusercontent.com/7459881/49382934-3d36b180-f6dd-11e8-978b-119574dec98d.png)

tiliev:
@lalo-mx, currently StackBlitz has some issues with caching npm package versions (https://github.com/stackblitz/core/issues/624). This is why the dependencies cannot be resolved.
Each sample from this repo can be found in our documentation - https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid.html, where it is optimized, tested and runnable in StackBlitz. Just use the `View on StackBlitz` buttons. We strongly recommend this approach instead of importing the whole project.

lalo-mx:
The documentation is usually useful but is not up to date as fast as the releases.  The StackBlitz is a fast way of testing the new features and fixes.

zdrawku:
Also, you can use the [angular-samples](https://github.com/IgniteUI/igniteui-angular-samples/releases) versions that we tag, we are going to update them regularly based on the included new samples and version updates.


Steps to execute:
1. Download the package 
2. `npm i`

3. `npm run start`

PR #1180, opened 2019-06-03
Update to 7.3.3

zdrawku:
Updating igniteui-angular to 7.3.3


hanastasov:
> err
> The samples cannot be opened in StackBlitz with this version of igniteui-angular.
> 
> There is no such error with the previous version

Seems like issue is related to npm's throttling/caching. We have had such issues before( see https://github.com/IgniteUI/igniteui-angular-samples/issues/782) and as @tiliev [stated](https://github.com/stackblitz/core/issues/837#issuecomment-452744028), we fixed it by publishing a new version.

zdrawku:
Related to https://github.com/stackblitz/core/issues/624

Issue #1177, opened 2019-05-31
Add a sample demonstrating tree grid remote paging


gedinakova:
### Description  
Add a sample demonstrating tree grid remote paging for root records.

PR #1181, opened 2019-06-04
Add tree grid remote paging sample

DiyanDimitrov:
#1177


src/app/tree-grid/tree-grid-remote-paging-sample/tree-grid-remote-paging-sample.component.html
```diff
@@ -0,0 +1,33 @@
+<igx-tree-grid #treeGrid [data]="data | async" childDataKey="Content" expansionDepth="0" width="100%" height="550px"
```
tachojelev: We can add an additional margin, so the grid does not get cut off on the top and on the bottom:
image

DiyanDimitrov: Done.

PR #1206, opened 2019-06-19
Update data-chart-samples-module.ts

mddifilippo89:
There was only one issue with this branch and it had to to with the namespace change to trendline


mddifilippo89:
@zdrawku @Lipata 
The checks failed because of the issue I was trying to fix with the namespace change. 

[21:10:55] Error: Cannot find module 'igniteui-angular-charts/ES5/igx-data-chart-category-trendline-module'

Lipata:
@mddifilippo89 maybe it is not the only place that is needed to be renamed? I can check this and fix it, but are there some other module name changes for the DV components?

mddifilippo89:
@Lipata 

Check now, 
I found one more place left with it. In live-editing\configs

PR #1225, opened 2019-06-27
Push all of my Angular samples to the browser.

agoldenbaum:


zdrawku:
@agoldenbaum there is no need to increase the travis build time, we do not rely on travis for the time being, we use Azure instead, so if the build passes on Azure, you can merge it.

zdrawku:
Merging this PR, Azure is passing

PR #1228, opened 2019-06-28
Add Mutli Cell Selection styling samples

ViktorSlavov:


ViktorSlavov:
@hanastasov 
Thanks for reviewing! I agree that, especially in the HGrid case, there is a bit too much markup. However, the aim of these samples is to showcase the theming and I want them to be as close as possible to the base samples.
The template refactoring could be done in a separate PR though, thanks for pointing it out!

Issue #1220, opened 2019-06-27
There is scrollbar visible on FInjsDemos


hanastasov:
### Description  
Describe the issue.

### Steps to reproduce  

1. Change your Display settings to 125%:
https://us-api.asm.skype.com/v1/objects/0-eus-d9-ebaafc7f0de5427f923f8fb570ef8e40/views/imgpsh_mobile_save

2. Go to https://www.infragistics.com/products/ignite-ui-angular/angular/components/treegrid/live_data.html

### Result  
Scrollbar around the iframe is visible

### Expected result  
No Scrollbar to appear

PR #1235, opened 2019-06-28
Diminish height for demos (7.3.x)

HristoP96:
Related to  #1220


HristoP96:
This fix should be applied also for the IgxGrid, because it reproduces the same behavior on zoom

PR #1237, opened 2019-06-28
Add polyfill template for IE

### LINKED ISSUE #1226: Stackblitz samples won't load in IE11 with Angular 8
### Description  
Loading samples with Angular 8 dependencies **(not dev)** doesn't load under IE11.
This is probably due to the conditional polyfills we rely on (since angluar/cli@7.3) but the dev deps in SB are not up-to-date:
```json
"devDependencies": {
    "@angular-devkit/build-angular": "~0.10.0",
    "@angular/cli": "~7.0.2",
    "@angular/compiler-cli": "~7.0.0",
    ...
```
This is quite likely what's causing other similar issues (https://github.com/stackblitz/core/issues/1004) as we only have API for 'dependencies'.

@HristoP96 You might be able to work around that by adding the missing poly-s just for IE generation.

### Steps to reproduce  

1. Open the main Data Grid page
2. Click the View on Stackblitz button
3. Observe the console

### Result  
Due to some missing poly the sample fails to load and an error is thrown:
Error: Object doesn't support property or method 'fromCodePoint'

### Expected result  
The sample should load

### Attachments  
image

HristoP96:
Closes #1226

### DIFF
--- a/live-editing/generators/SharedAssetsGenerator.ts
+++ b/live-editing/generators/SharedAssetsGenerator.ts
@@ -17,7 +17,7 @@ const ANGULAR_JSON_TEMPLATE_CSS_SUPPORT_PATH =
 const MAIN_TS_FILE_PATH = path.join(__dirname, "../templates/main.ts.template");
 const APP_COMPONENT_SCSS_PATH = path.join(__dirname, "../../src/app/app.component.scss");
 const APP_COMPONENT_TS_PATH = path.join(__dirname, "../../src/app/app.component.ts");
-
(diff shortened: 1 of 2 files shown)


HristoP96:
> Everything works as expected on IE 11, Goot job! Although for Edge this is not the case. The buttons are not enabled. Image and error below:
> image

Yes, this is due Edge's security that does not allow localhost to fetch data from a server without configured cors options. This can be omitted if you run fiddler. We must investigate this further why this behavior is prevented with the fiddler

Issue #1264, opened 2019-07-05
Fix broken samples and sample configs


HristoP96:
### Description  
This issue is regarding the following samples:
- Missing modules in config file:
   - **DataChartAxisSharingComponent**
   - **DataChartTypeFinancialSeriesComponent**
   - **DataChartTypeScatterAreaSeriesComponent**
   - **DataChartTypeScatterSeriesComponent**
   - **DataChartTypeStackedSeriesComponent**
   - **DataChartSeriesTrendlinesComponent**
   - **DataChartTypeValueOverlayComponent**
   - **FinancialChartHighFrequencyComponent**

- No config and has StackBlitz button:
  - **MapDisplayImageryHeatTilesComponent**

- Wrong iframe path:
  - **SpreadsheetCondFormatComponent**

Remove caret from **igniteui-angular** version

HristoP96:
Done

PR #1265, opened 2019-07-05
Fix configs, rename wrong set spreadsheet folder, add map config, update igniteui-anguilar

HristoP96:
Cloese #1264


zdrawku:
Bump the version to 8.0.2 and we are ready

PR #1272, opened 2019-07-10
feat(input-group): add input group styling example

ViktorSlavov:


src/app/data-entries/input-group/input-group-styling/input-group-styling.component.scss
```diff
@@ -0,0 +1,32 @@
+@import '~igniteui-angular/lib/core/styles/themes/index';
+
+$purple: #9E379F;
+$blue: #61AEDB;
```
wnvko: It's not a big deal but I would put primary (blue) color above secondary (purple) color.

PR #1278, opened 2019-07-11
Add a sample with transactions

jackofdiamond5:


src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    public price: string;
+    public addDisabled;
+    public deleteDisabled;
+
+    constructor(public transactions: IgxTransactionService<Transaction, State>) {
+        this.wishlist = WISHLIST;
+    }
+
+    public onAdd(event): void {
+        const person: IItem = { name: "Yacht", price: "A lot!" };
```
dafo: This would be better if named "item" probably.

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        }, this.wishlist[1]);
+
+        /** visualization */
+        this.editDisabled = true;
+        this.name = this.wishlist[1].name;
+        this.price = newPrice;
+    }
+
+    public onDelete(event): void {
+        /** visualization */
```
dafo: The /** visualization */ may be below the transaction itself like in the rest of the methods.

live-editing/configs/TransactionsConfigGenerator.ts
```diff
@@ -0,0 +1,32 @@
+import {
+    IgxCardModule, IgxChipsModule, IgxExpansionPanelModule,
+    IgxIconModule, IgxListModule, IgxTransactionService
+} from "igniteui-angular";
+import {
+    TransactionBaseComponent
```
wnvko: Import this from igniteui-angular

src/app/services/transaction/pipes/transaction-base.pipe.ts
```diff
@@ -0,0 +1,42 @@
...
+export class TransactionBasePipe implements PipeTransform {
+    // inject the transaction service
+    constructor(public transactions: IgxTransactionService<Transaction, State>) { }
+
+    public transform(data: WishlistItem[]) {
+        // the pipe should NOT operate on the original dataset
+        // we create a copy of the original data and then use it for visualization only
+        const _data = [...data];
+        const states = this.transactions.getAggregatedChanges(false);
+        // iterate over all pending transactions
```
wnvko: we are iterating the states, not the transactions

src/app/services/transaction/pipes/transaction-base.pipe.ts
```diff
@@ -0,0 +1,42 @@
...
+        for (const state of states) {
+            // depending on the type of transaction either:
```
wnvko: depending on the type of the state. You may have add transaction and then update transaction for same id. The type of the state will be add in this case

src/app/services/transaction/pipes/transaction-base.pipe.ts
```diff
@@ -0,0 +1,42 @@
...
+            switch (state.type) {
+                case TransactionType.ADD:
+                    // push the newValue property of the current ADD transaction
+                    _data.push(state.newValue);
```
wnvko: here also change transaction with state

src/app/services/transaction/pipes/transaction-base.pipe.ts
```diff
@@ -0,0 +1,42 @@
...
+                    break;
+                // we do not directly operate on records, we just style them
+                // the record will be deleted once the transaction is committed
```
wnvko: add something about why we do not change the data row itself - we should not change the original data

src/app/services/transaction/transaction-base/transaction-base.component.html
```diff
@@ -0,0 +1,64 @@
...
+                    </igx-list-item>
+                </ng-template>
+            </igx-list>
+        </igx-card>
+    </div>
+    <div class="dynamic-list">
+        <igx-card>
+            <igx-list>
+                <igx-list-item [isHeader]="true">Wishlist (with transactions)</igx-list-item>
+                <ng-template ngFor [ngForOf]="this.wishlist | transactionBasePipe" let-item>
```
wnvko: add short comment about why we are using the pipe here

src/app/services/transaction/transaction-base/transaction-base.component.html
```diff
@@ -0,0 +1,64 @@
...
+    </div>
+</div>
+<div class="options-wrap">
+    <div class="options">
+        <igx-expansion-panel [collapsed]="false">
+            <igx-expansion-panel-header iconPosition="none" [disabled]="true">
+                <button #add igxButton (click)="onAdd($event)" [disabled]="addDisabled" class="add">Add
+                    New</button>
+                <button #edit igxButton (click)="onEdit($event)" class="edit"
+                    [disabled]="editDisabled || wishlist.length < 2">Edit
```
wnvko: wishlist.length < 2 this is not necessary, is it?

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -0,0 +1,172 @@
...
+    styleUrls: ["./transaction-base.component.scss"],
+    templateUrl: "transaction-base.component.html"
+})
+export class TransactionBaseComponent {
+    public wishlist: WishlistItem[];
+    public addDisabled: boolean;
+    public deleteDisabled: boolean;
+    public editDisabled: boolean;
+
+    // inect the transaction service
```
wnvko: inject :)

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    /*
+     * For each type of action that we want to perform we can add different types of transactions.
+     * The below methods - onAdd, onEdit, onDelete demonstrate the different types of transactions
+     * and what kind of data we normally pass to each of them.
+     *
+     * Each transaction needs to have id, type and a newValue properties.
+     * ADD and DELETE transactions should have unique IDs.
+     * There can be multiple UPDATE transactions with the same ID,
+     * since there can be multiple consecutive changes on a single element.
```
wnvko: ... on a single data row - not element

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -0,0 +1,172 @@
...
+    */
+
+    /**
+     * Create an 'ADD' transaction.
+     */
+    public onAdd(event): void {
+        // the item that will be added, it could be anything
+        // it is a good idea for it to have a unique 'id' property
```
wnvko: it **must** have unique ID, otherwise exception will be thrown. And the reason for this is not to find it easier, so you could remove the third line of the comment.

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -0,0 +1,172 @@
...
+     * Each transaction needs to have id, type and a newValue properties.
+     * ADD and DELETE transactions should have unique IDs.
+     * There can be multiple UPDATE transactions with the same ID,
+     * since there can be multiple consecutive changes on a single element.
+    */
+
+    /**
+     * Create an 'ADD' transaction.
+     */
+    public onAdd(event): void {
```
wnvko: The interesting thing about ADD transaction is that you do not provide `recordRef`, as you do not have one. Add comment about this.

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Clear all pending transactions.
+     */
+    public onClear(event): void {
+        /** visualization  */
+        if (this.transactions.getTransactionLog().length !== 0) {
+            this.reset();
+        }
+        /** */
```
wnvko: leftover?

PR #1287, opened 2019-07-15
Add styling sample for column hiding

tachojelev:
IgniteUI/igniteui-docfx#1275


src/app/grid/grid-column-hiding-toolbar-style/grid-column-hiding-toolbar-style.component.scss
```diff
@@ -0,0 +1,80 @@
...
+
+/*BASIC THEMING*/
+
+$custom-column-hiding-theme: igx-column-hiding-theme(
+    $background-color: steelblue,
+    $title-color: gold
+);
+$custom-button: igx-button-theme($flat-text-color: gold, $disabled-color: black);
+
+/*INCLUSION by DEFAULT ViewEncapsulation strategy*/
```
HristoP96: It would be better if the inclusion is added after all the styling approaches, so that only with the removing of the comments, the themes could be applied

src/app/grid/grid-column-hiding-toolbar-style/grid-column-hiding-toolbar-style.component.scss
```diff
@@ -0,0 +1,80 @@
...
+//             igx-color:("secondary", 500)
+//         ),
+//         disabled-color:(
+//             igx-color:("primary", 700)
+//         )
+//     )
+// );
+// 
+// $custom-dark-schema: extend($dark-schema,(
+//     igx-grid-column-hiding: $custom-column-hiding-schema,
```
HristoP96: the property name is ` igx-column-hiding`

PR #1299, opened 2019-07-17
Add banner styling sample

IvayloG:
Related https://github.com/IgniteUI/igniteui-docfx/pull/1384


IvayloG:
Although adding button theme section is not the topic, no doubt the sample will look better with styled buttons overall.

Issue #1266, opened 2019-07-05
Drag and Drop new samples


mpavlinov:
### Description  
Add new samples for the Drag and Drop

- [x] Icons re-arrangement - [Archive.zip](https://github.com/IgniteUI/igniteui-angular-samples/files/3363240/Archive.zip)
- [x] ~~Calendar~~
- [x] Dialog
- [x] E-mail
- [x] Kanban board
- [x] List reorder

https://github.com/IgniteUI/igniteui-angular/issues/5319

PR #1300, opened 2019-07-17
Adding the List Reorder drag and drop sample

PavlovVasil:
Related: https://github.com/IgniteUI/igniteui-angular-samples/issues/1266


StefanIvanov:
List item start drag should be triggered only from the drag icon on the right side of the list item, now it is triggered by the whole item.

PR #1310, opened 2019-07-18
Fix issues regarding MRL configurator sample.

### LINKED ISSUE #1309: MRL Layout Configurator sample issues
### Description  
The following sample has some issues: https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/multi_row_layout.html#layout-configurator

On Chrome:
1) Cannot drag column chips into layout areas.
dragDrop

On IE:
1) When layout is focused its styles are not applied correctly:
image
2) Container shows no horizontal scrollbar, so not the whole layout is visible.

Chrome shows scrollbar:
image

On IE, no scrollbar:
image

3) Cannot delete a column from the layout.
When you select a column from the layout and hit Delete key nothing happens.
The column should be removed from the layout.

skrustev:
Closes #1309


src/app/grid/grid-multi-row-layout-configuration/grid-multi-row-layout-configuration.component.scss
```diff
@@ -20,7 +20,7 @@
 }
 
 .layoutListContainer {
-    width: fit-content;
+    display: fit-content;
```
dkamburov: There is no `fit-content` value for the `display` property. Probably should be removed ?

MayaKirova:
@skrustev Other issues seem to be fixes, only the last one for IE remains:

Cannot delete a column from the layout.
When you select a column from the layout and hit Delete key nothing happens.
The column should be removed from the layout.

PR #1326, opened 2019-07-22
Update transactions demo's styling

### LINKED ISSUE #1303: Update transaction demo
### Description  
Update transaction demo in How to use the Transaction Service topic with the design, provided by UX team.

### Attachments  
new-design

jackofdiamond5:
Closes #1303


Lipata:
Just put some margin-right to the static list at the top right, because in StackBlitz the list is glued to the right border.

IvayloG:
From a user perspective, it will be nice either if the buttons are disabled when there are no more user interactions available OR a way to reset the sample in its initial state. Optionally, Commit and Clear Transactions buttons could be Enabled only when there is an actual pending transaction.

IvayloG:
Looks nice. One small thing - there is a bit of excess currency formatting. Can you remove the rightmost dollar sign for both Transaction Log and Data Items and using the ones on the left? 

image

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -13,6 +13,9 @@ export class TransactionBaseComponent {
     public deleteDisabled: boolean;
     public editDisabled: boolean;
 
+    public applyDelete: boolean;
+    public applyEdit: boolean;
+
     // inject the transaction service
```
damyanpetev:
```suggestion
    /**
     * @param transactions Injected Transaction Service.
     */
```

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -94,77 +99,29 @@ export class TransactionBaseComponent {
...
+    public onCommit(): void {
         // the commit function expects the dataset as its parameter
         this.transactions.commit(this.wishlist);
+
+        /** visualization */
+        this.applyDelete = false;
+        this.applyEdit = false;
     }
 
     /*
```
damyanpetev:
```suggestion
    /*********************************************************************************************
```

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -45,45 +50,47 @@ export class TransactionBaseComponent {
...
 
     /**
-     * Create a 'DELETE' transaction.
+     * Create a `DELETE` transaction.
      */
-    public onDelete(event): void {
-        // there cannot be two or more DELETE transactions with the same id
-        // the 'newValue' property should be set to null since we do not change any values,
+    public onDelete(): void {
+        // there cannot be more than one `DELETE` transactions with the same `id`
```
damyanpetev:
```suggestion
// After a `DELETE` transaction, no further changes should be made for the same `id`
```

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -45,45 +50,47 @@ export class TransactionBaseComponent {
...
+        // the `newValue` property should be set to `null` since we do not change any values,
         // we just delete the entire record
         this.transactions.add({
```
damyanpetev: split the `deleteTransaction` into a typed variable above for readability.

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -94,77 +101,29 @@ export class TransactionBaseComponent {
     /**
      * Commit all pending transactions.
      */
-    public onCommit(event): void {
-        // the commit function expects the dataset as its parameter
+    public onCommit(): void {
+        // the `commit` function expects the dataset as its parameter
```
damyanpetev:
```suggestion
        // the `commit` function expects the original data array as its parameter
```

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -13,29 +13,34 @@ export class TransactionBaseComponent {
     public deleteDisabled: boolean;
```
damyanpetev: `addDisabled` can and probably should be replaced with a check against `state` and the data by id. It's just for demo purposes anyway.

damyanpetev: `deleteDisabled` should be a check like so:
```typescript
/** Check item for unsaved deletion */
public isDeleted(id) {
    const state = this.transactions.getState(id);
    return state && state.type === TransactionType.DELETE;
}
This can be used for the class application (since it's actually useful logic in a real scenario) and for the edit button check,
```

damyanpetev: Same applies as `isEdited` check.

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -13,29 +13,34 @@ export class TransactionBaseComponent {
...
     public editDisabled: boolean;
 
-    // inject the transaction service
+    public applyDelete: boolean;
+    public applyEdit: boolean;
```
damyanpetev: TBH, the checks are arbitrary for the demo, I'd be okay if they are visibly so:
```html
<igx-icon igxListAction (click)="onEdit()" *ngIf="item.id === 1 && item.price !== '$999'">edit</igx-icon>
```
And just drop the entire state variable thing that's sprinkled through the code.

src/app/services/transaction/transaction-base/transaction-base.component.ts
```diff
@@ -94,77 +101,29 @@ export class TransactionBaseComponent {
...
-                // get the color that corresponds to the state's type
-                return this.getColor(transaction);
-            }
-        }
-
-        return null;
-    }
-
     /**
      * Map the items in the transaction log with the items in the wishlist.
```
damyanpetev:
```
/* Get transactions, merge with wishlist data for display purposes */
```

PR #1355, opened 2019-07-29
select topic samples update

IvayloG:
Add new samples related to the igxSelect topics update. 
Related https://github.com/IgniteUI/igniteui-docfx/issues/1417


src/app/data-entries/select/select-form/select-form.component.html
```diff
@@ -0,0 +1,15 @@
+<form>
+    <div class="select-wrapper">
+        <igx-select #selectModel="ngModel" [(ngModel)]="selected" [ngModelOptions]="{standalone: true}" required>
```
damyanpetev: Wouldn't mind this using a name inside the form TBH:
```suggestion
        <igx-select #selectModel="ngModel" [(ngModel)]="selected" name="fruit" required>
```

src/app/data-entries/select/select-input-directives/select-input-directives.html
```diff
@@ -0,0 +1,14 @@
+<div class="select-wrapper">
+    <igx-select [(ngModel)]="selected">
+        <label igxLabel>Pick a fruit</label>
+        <igx-prefix>
+            <igx-icon color="green" style="width:48px; height:32px;">BIO</igx-icon>
```
damyanpetev: Ooops, this shouldn't really be an igx-icon:
```suggestion
            <span class="bio-prefix">BIO</span>
```
And then you'd need an extra style (see below). This will obv need to be reflected in the docs snippet.

src/app/data-entries/select/select-input-directives/select-input-directives.scss
```diff
@@ -0,0 +1,5 @@
+.select-wrapper {
+    display: flex;
+    padding-top: 160px;
+    padding-left: 40px;
+}
```
damyanpetev:
```suggestion
}
.bio-prefix {
    color: green;
    font-weight: 500;
    font-size: 1.5rem;
}
```

PR #1373, opened 2019-08-05
Adds sample for custom summaries based on selection

### LINKED ISSUE #1170: Selection based summaries sample
Add "Selection based summaries" as a new section with a sample scenario to the existing [grid selection topic](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/selection.html#grid-row-selection) after the "Row Selection" section and before "Setup". The height of the row that is added at the bottom matches the summary row display density for the configured density setting and keeps a padding on the right that also matches the one of a summary cell.

Selection Aggregates 1
Selection Aggregates 2

3phase:
Closes #1170


3phase:
> * I think there should be some initial selection, because right now it's not clearly visible what this sample does.
> * Also you should change "aggregate" with "summaries", because we already have introduced "summaries" as a feature and a term to the customers in the documentation. Introducing a new term will confuse them.
> * Display density doesn't have effect on the custom summaries.

Done

mpavlinov:
@StefanIvanov Yes. We will add them.

PR #1380, opened 2019-08-07
Drag and Drop Email-Sample

3phase:
Related: #1266


mpavlinov:
@3phase There is a problem in this PR. The merge commit was not good resulting in 90 modified files.

PR #1391, opened 2019-08-13
Drag and Drop Email-Sample

3phase:
Related: #1266
Fixes faulty branch #1380


mpavlinov:
**Note:** There are new options in this sample so a 8.2.0 beta is needed for build to pass.

PR #1458, opened 2019-09-05
Fixes issues with custom selection and removes redundant code

### LINKED ISSUE #1457: Custom Selection Won't Select Single Elements Properly
### Description  
In the custom selection summaries sample, when selecting multiple single cells using Ctrl, the result of the selection is wrong

### Steps to reproduce  

1. Open the custom selection sample
2. Select a cell
3. Using ctrl + mouse click, select another cell from a different column
4. Observe how the summary changes

### Result  
The new result would overwrite the existing one, no matter that both of them should be applied to the summary.

### Expected result  
Both of the cells have to have an impact on the summary when both of them are selected.

3phase:
Closes #1457 
Test w/ IgniteUI/igniteui-angular#5735 or after merging it (tagged as beta 4)


3phase:
After discussions with @ddincheva and @mpavlinov, we decided to replace the custom-built footer with the igx-grid-footer component, which turned out to be blocked by the IgniteUI/igniteui-angular#5839 issue.

3phase:
The issue IgniteUI/igniteui-angular#5839 appeared to be applicable to the 8.2.0 beta versions only, so it can be tested now.

Issue #1377, opened 2019-08-07
Show Sparkline in all Primary Grids


zdrawku:
### Description  

See [this example](https://www.infragistics.com/products/ignite-ui-angular/angular/components/sparkline.html#sparkline-in-data-grid) of Sparkline within a Grid.

Add Sparklines for the following Grids:

- [x] Boston Marathon - https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/grid.html
- [x] TreeGrid - https://www.infragistics.com/products/ignite-ui-angular/angular/components/treegrid/tree_grid.html
- [x] HierarchicalGrid - https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/hierarchical_grid.html
- [x] Business Propeller - https://www.infragistics.com/products/ignite-ui-angular
- [x] FinJS and Crypto Grids - https://www.infragistics.com/resources/sample-applications

Also, update the documentation to address this change, with explanations on how to configure it.

zdrawku:
We can remove the last point of the list, I am adding chart logic in the FinJS grid with [this PR](https://github.com/IgniteUI/igniteui-angular-samples/pull/1414), and as for the Crypto Grid, this will be handled later on.

zdrawku:
Done

PR #1462, opened 2019-09-09
Add sparkline column to grid demos

martinesoff:
Related to #1377


HristoP96:
Everything seems to be working in Chrome EDge Ffox and ie11, 

When the respective configs are updated with the necessary files, the PR will be ready for merge

PR #1467, opened 2019-09-13
fix(card): fix horizontal cad sample in IE11

### LINKED ISSUE #1319: [IE] Horizontal oriented card samples are misaligned
### Description  
Horizontal oriented card samples are misaligned.

NOTES:
- The issue is reproduced in IE only.
- When the sample is opened in stackblitz the issue is not reproduced, since it seems that stackblitz removed the flex related styles.

- **RELATED TO** https://github.com/IgniteUI/igniteui-docfx/issues/1541
### Steps to reproduce  

1. Navigate to the following link:
https://www.infragistics.com//angularsite/components/card.html#horizontal-layout

### Result  
Right portion of the card with the actions is not visible

### Expected result  
Right portion of the card with the actions is visible.

### Attachments  

image

desig9stein:
Closes #1319

### DIFF
--- a/src/app/layouts/card/card-sample-2/card-sample-2.component.html
+++ b/src/app/layouts/card/card-sample-2/card-sample-2.component.html
@@ -1,6 +1,6 @@
 <div class="sample-column card-wrapper">
     <igx-card type="outlined" [horizontal]="horizontal">
-        <div igxLayout igxLayoutDir="column" igxFlex igxFlexGrow="1">
+        <div class="h-sample-column">
             <igx-card-header>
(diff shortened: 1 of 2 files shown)


tachojelev:
Since this PR includes new styling changes and the usage of the igxLayout and igxFlex directives is removed, the documentation should be updated as well.
Issue: https://github.com/IgniteUI/igniteui-docfx/issues/1541

zdrawku:
In that case, this PR should not be merged until there is a documentation PR for the same. Then we can merge both PRs simultaneously

PR #1481, opened 2019-09-19
[DO NOT MERGE] Clear selection when filtering is done FinJs

### LINKED ISSUE #1480: Chart data is stacking the newly filtered items
### Description  
The chart should populate only the newly selected data, instead of adding it to the old one.

### Steps to reproduce  

1. Open
2. Open the filter dialog for Category column
3. Clear all filters and select "Agriculture"
4. Press ok, the data will be filtered to show only Agriculture records
5. Select 3 rows
6. Click on the Chart button (the last button of the Button group)

### Result  
The old data still persists, and the new data is added to it

### Expected result  
To show me the data only of the newly selected 3 rows.

### Attachments  
image

hanastasov:
Closes #1480


zdrawku:
I agree with @nrobakova and @hanastasov, this is the expected behavior.

Issue #1494, opened 2019-09-26
Empty rows in the transactions grid


DiyanDimitrov:
### Description  
Empty rows in the transactions grid when editing cells.

### Steps to reproduce  

1. Open https://staging.infragistics.com/products/ignite-ui-angular/angular/components/treegrid/batch_editing.html#demo
2. Edit a cell value.
3. Click the Commit button.

### Result  
The transactions grid has an empty row.

### Expected result  
The transaction grid should display the Update transaction.

### Attachments   
![image](https://user-images.githubusercontent.com/43128948/65672125-32288580-e051-11e9-9f03-cec58a124eb7.png)

PR #1495, opened 2019-09-26
Put treegrid in component with transactions

wnvko:
`IgxTreeGrid` batch editing sample has one more grid in the transactions confirmation dialog. As the tree grid has `IgxHierarchicalTransactionService` provided same service is used in the confirmation dialog grid. This breaks the grid in the dialog. To solve the issue we are putting the tree grid in separate component - `TreeGridWithTransactionsComponent` where the transactions service is provided.

Closes: #1494


live-editing/configs/TreeGridConfigGenerator.ts
```diff
@@ -26,6 +26,7 @@ import {
 import { IgxSparklineCoreModule} from "igniteui-angular-charts/ES5/igx-sparkline-core-module";
 import { IgxSparklineModule} from "igniteui-angular-charts/ES5/igx-sparkline-module";
 import { TreeGridBatchEditingSampleComponent } from "../../src/app/tree-grid/tree-grid-batch-editing/tree-grid-batch-editing-sample.component";
+import { TreeGridWithTransactionsComponent } from "../../src/app/tree-grid/tree-grid-batch-editing/tree-grid-with-transactions.component"
```
ViktorSlavov: Semi column is missing at the end of the line.

live-editing/configs/TreeGridConfigGenerator.ts
```diff
@@ -186,8 +187,9 @@ export class TreeGridConfigGenerator implements IConfigGenerator {
         configs.push(new Config({
             additionalFiles: ["/src/app/data/utils.ts", "/src/app/tree-grid/data/employees-flat.ts"],
```
ViktorSlavov: The patch to the `TreeGridWithTransactionsComponent` should be added, so the StackBlitz sample properly works:
```suggestion
             additionalFiles: ["/src/app/data/utils.ts", "/src/app/tree-grid/data/employees-flat.ts", "/src/app/tree-grid/tree-grid-batch-editing/tree-grid-with-transactions.component.ts"],
```

PR #1501, opened 2019-10-01
Simplify select groups demo

### LINKED ISSUE #1419: Simplify select groups demo
### Description  
The Select With Groups demo at https://www.infragistics.com/products/ignite-ui-angular/angular/components/select.html#select-with-groups is a bit more complicated than needed due to having a data array with mismatched properties. That requires double ngFor which can totally be avoided:
https://stackblitz.com/edit/select-groups?file=src/app/select-sample-2/select-sample-2.component.html

IvayloG:
Closes #1419


damyanpetev:
@IvayloG rebase against the latest branch state please :)

PR #1505, opened 2019-10-01
Fix grid persistence sample - master

tachojelev:
Closes IgniteUI/igniteui-angular#5842

### DIFF
--- a/src/app/grid/grid-save-state/grid-state.component.html
+++ b/src/app/grid/grid-save-state/grid-state.component.html
@@ -38,7 +38,7 @@
     <igx-grid [id]="gridId" #grid1 igxState [data]="localData" [primaryKey]="'EmployeeID'" width="950px" height="550px"
         [rowEditable]="true" [allowFiltering]="true" [allowAdvancedFiltering]="true" [rowSelectable]="true"
         [paging]="true" [showToolbar]="true" [columnPinning]="true" [columnHiding]="true"
-        [filterMode]="'excelStyleFilter'">
+        [filterMode]="'excelStyleFilter'" [displayDensity]="'compact'">
         <igx-column *ngFor="let c of columns" [sortable]="true" [movable]="true" [filterable]="true"
             [field]="c.field" [header]="c.header" [dataType]="c.dataType" [pinned]="c.pinned" [hidden]="c.hidden"></igx-column>
     </igx-grid>
--- a/src/app/grid/grid-save-state/grid-state.component.html
+++ b/src/app/grid/grid-save-state/grid-state.component.html
@@ -15,6 +15,9 @@
                 <igx-switch [(ngModel)]="state.filtering" (change)="onChange($event, 'filtering')">Filtering</igx-switch>
             </div>
             <div class="control-item">
+                <igx-switch [(ngModel)]="state.advancedFiltering" (change)="onChange($event, 'advancedFiltering')">Advanced Filtering</igx-switch>
+            </div>
+            <div class="control-item">
                 <igx-switch [(ngModel)]="state.paging" (change)="onChange($event, 'paging')">Paging</igx-switch>
             </div>
             <div class="control-item">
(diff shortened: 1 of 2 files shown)


src/app/grid/grid-save-state/state.directive.ts
```diff
@@ -215,5 +209,23 @@ export class IgxGridStateDirective implements AfterViewInit {
         });
         return gridColumns;
     }
+
+    private stringifyCallback(key: string, val: any) {
+        const expr = val as IFilteringExpression;
+        if (expr && expr.searchVal instanceof Set) {
+            expr.searchVal = Array.from(expr.searchVal);
+            return expr;
+        }
```
DiyanDimitrov: Instead of modifying the expression, you should check the `key` and `val` parameters of the function:
```ts
        if (key === "searchVal" && val instanceof Set) {
            return Array.from(val);
        }
```

The same should be applied to the parseCallback too.

src/app/grid/grid-save-state/grid-state.component.html
```diff
@@ -33,8 +36,9 @@
     </div>
 
     <igx-grid [id]="gridId" #grid1 igxState [data]="localData" [primaryKey]="'EmployeeID'" width="950px" height="550px"
-        [rowEditable]="true" [allowFiltering]="true" [rowSelectable]="true"
-        [paging]="true" [columnPinning]="true">
+        [rowEditable]="true" [allowFiltering]="true" [allowAdvancedFiltering]="true" [rowSelectable]="true"
+        [paging]="true" [showToolbar]="true" [columnPinning]="true" [columnHiding]="true"
+        [filterMode]="'excelStyleFilter'" [displayDensity]="'compact'">
```
DiyanDimitrov: `[displayDensity]="'compact'"` looks a bit tiny, could you change that to cosy? You will also need to update the sample height in the topic.

tachojelev: Done

src/app/grid/grid-save-state/state.directive.ts
```diff
@@ -85,72 +93,51 @@ export class IgxGridStateDirective implements AfterViewInit {
...
-          this.grid.selectRows(this.gridState.selection);
+            this.grid.selectRows(this.gridState.selection);
         }
     }
 
     public storeState(action: string, args: any) {
         if (this[action]) {
             const actionKey = action + "-" + this.grid.id;
-            window.localStorage.setItem(actionKey, JSON.stringify(args));
+            window.localStorage.setItem(actionKey, JSON.stringify(args, this.stringifyCallback));
```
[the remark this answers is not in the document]
tachojelev: Done

src/app/grid/grid-save-state/state.directive.ts
tachojelev: Done

tachojelev: Done

src/app/grid/grid-save-state/state.directive.ts
```diff
@@ -85,72 +95,56 @@ export class IgxGridStateDirective implements AfterViewInit {
...
-    public storeState(action: string, args: any) {
+    public storeState(action: string, args: any, stringifyCb?: (key, val) => any) {
         if (this[action]) {
             const actionKey = action + "-" + this.grid.id;
-            window.localStorage.setItem(actionKey, JSON.stringify(args));
+            if (stringifyCb) {
+                window.localStorage.setItem(actionKey, JSON.stringify(args, stringifyCb));
+            } else {
+                window.localStorage.setItem(actionKey, JSON.stringify(args));
+            }
```
DiyanDimitrov: There is no need to check `stingifyCb`, you can always call `window.localStorage.setItem(actionKey, JSON.stringify(args, stringifyCb));`. Same is valid for `parseCb` below.

tachojelev: Done

src/app/grid/grid-save-state/state.directive.ts
```diff
@@ -55,27 +60,32 @@ export class IgxGridStateDirective implements AfterViewInit {
...
         this.storeState("sorting", sortingState);
 
-        const filteringState = { filtering: this.grid.filteringExpressionsTree};
-        this.storeState("filtering", filteringState);
+        const filteringState = { filtering: this.grid.filteringExpressionsTree };
+        this.storeState("filtering", filteringState, this.stringifyCallback);
 
-        const selectionState = {selection: this.grid.selectedRows()};
+        const advancedFilteringState = { advancedFiltering: this.grid.advancedFilteringExpressionsTree };
+        this.storeState("advancedFiltering", advancedFilteringState, this.stringifyCallback);
```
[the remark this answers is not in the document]
tachojelev: Done

Issue #1556, opened 2019-11-01
Update Remote Filtering and Sorting Samples


DiyanDimitrov:
Related to IgniteUI/igniteui-angular#4356

PR #1558, opened 2019-11-01
Update grid remote sorting/filtering samples - vNext

tachojelev:
#1556


DiyanDimitrov:
I have noticed an existing issue in the Grid Remote Filtering sample. When you scroll to bottom and then filter the ProductName column with contains 'c', 'No records found' message is displayed, but actually there are records that match the filter criteria.

PR #1569, opened 2019-11-05
Update vNext to igniteui-angular 9.0.0-beta.2

DiyanDimitrov:


ChronosSF:
@simeonoff , shouldn't there be another way to style the buttons in the toolbar themes?

PR #1594, opened 2019-11-20
fix(sample): Adding debounce when handling onDataPreLoad subscription…

### LINKED ISSUE #1555: Scrolling in the remote virtualization sample is not smooth
### Description  
Scrolling in the remote virtualization sample is not smooth.

### Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/virtualization.html#remote-virtualization-demo
2. Scroll down using the mouse wheel.

### Result  
After you stop scrolling row data is changed couple of times before the correct rows are visualized.

### Expected result  
The remote virtualization scrolling should be smooth.

MayaKirova:
…s. Refactoring the logic so that loading style is applied only when loading remote data (not when data is already cached).

Closes #1555


ChronosSF:
@DiyanDimitrov , please test the sample for the issue you experienced after Maya's changes.

PR #1599, opened 2019-11-22
Add paginator to remote paging samples

### LINKED ISSUE #1549: Optimize Grid Remote Paging Sample
### Description  
Currently our remote paging sample requires re-templating the pager and a lot of custom code https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/paging.html#remote-data.
The sample could be optimized if the `igx-paginator` component is used https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/paging.html#paginator-component.

hanastasov:
Closes #1549


hanastasov:
> 1. We should remove the second pager template from the grid remote paging sample. The sample should demonstrate only the remote paging.

The custom template demonstrates the same - the correct API to be used to achieve remote paging, which is essential.

PR #1620, opened 2019-11-29
Fix samples leaking css

### LINKED ISSUE #1619: Leaking drop down menu sample css
### Description  
Due to encapsulation: ViewEncapsulation.None, there leaking css rules to other samples.

### Steps to reproduce  

1. Navigate to http://localhost:4200/samples/data-entries/dropdown-menu
2. Navigate to http://localhost:4200/samples/data-entries/input-group-sample-6

### Result  
The following class is applied, that breaks the visuals.
    .label {
        margin: 15px;
    }

### Expected result  
No such class applied.

### Attachments  
image

IvayloG:
Closes #1619

For the $border-radius to be affected in the samples as well, it will require the following fix: https://github.com/IgniteUI/igniteui-angular/issues/6299


src/app/data-entries/dropdown/dropdown-menu/dropdown-menu.component.scss
```diff
@@ -1,31 +1,35 @@
-.drop-down-wrapper{
-    display: flex;
-    flex-direction: column;
-    padding-top: 5px;
-}
+::ng-deep {
```
damyanpetev: No need to `ng-deep` all the rules, just the ones that affect child elements not in the component view.

PR #1633, opened 2019-12-10
Update grid row reorder sample to include window scroll position.

skrustev:
Updating samples due to changes introduced with IgniteUI/igniteui-angular/issues/6180


ChronosSF:
@skrustev , pr this to vnext as well

PR #1635, opened 2019-12-12
Add header footer sample 

### LINKED ISSUE #1571: Create demo for igx-select with header and footer templates
### Description  
 [feature](https://github.com/IgniteUI/igniteui-angular/issues/5911)

Current Demos Dev sample:
image

IvayloG:
Closes #1571


Lipata:
It is not possible to deselect and already selected delivery method

valadzhov:
Assigning the task back to Ivaylo because there's inconsistent behaviour between the sample in the topic and the one in stackblitz (no style is being applied to the delivery method icons).

PR #1636, opened 2019-12-12
Changing hierarchical batch updating sample for new hierarchical transactions injection

ChronosSF:
Changed how the batch sample works because transactions of children are now split (fix for bugs https://github.com/IgniteUI/igniteui-angular/issues/6074 & https://github.com/IgniteUI/igniteui-angular/issues/5500 )


ChronosSF:
Requires https://github.com/IgniteUI/igniteui-angular/pull/6282 to appear in a beta.

PR #1637, opened 2019-12-12
Removing the dark overlay when loading

3phase:
Removes dark overlay when sorting remote filtering

### DIFF
--- a/src/app/grid/grid-remote-filtering-sample/remote-filtering-sample.component.scss
+++ b/src/app/grid/grid-remote-filtering-sample/remote-filtering-sample.component.scss
@@ -32,3 +32,12 @@
 .badge-left {
     float: left;
 }
+
+:host {
+    ::ng-deep {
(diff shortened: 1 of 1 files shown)


ChronosSF:
@kdinev, @zdrawku , this was requested for production asap. I guess we don't have to pr it to vNext as well in this case.

src/app/grid/grid-remote-filtering-sample/remote-filtering-sample.component.scss
```diff
@@ -31,4 +31,13 @@
 
 .badge-left {
     float: left;
+}
+
+:host {
+    ::ng-deep {
+        .igx-overlay__wrapper--modal {
+            background: none;
+            pointer-events: initial;
```
kdinev: You can still block all pointer events, just don't make the grid flicker (darkening it out). Also, this changes the sample, but the default in the grid still remains with the dark background. This should also be fixed in the product.

PR #1649, opened 2019-12-19
Enhancing Conditional styling in chart integration sample

### LINKED ISSUE #1648: Enhance conditional styling in chart integrations sample
Apply the following changes to the sample:

- Add clear format functionality
- Remove button rendering on right click
- Render data analysis button on scroll (horizontal and vertical), only when there is an active range selection
- Data analysis button should render only for active ranges
- On grid cell click deactivates all active ranges if present.
- Add support for editing of cells that are into a range with applied formatting (every cell edit should trigger formatting of the whole range, no matter if the range is active or not).

HristoP96:
Closes #1648


zdrawku:
@HristoP96 create the same PR for master

PR #1668, opened 2020-01-16
Providing chart integration directive

### LINKED ISSUE #1667: Chart Integration Directive
## Description  
Create a Directive that could be used from the igxGrid to easily set-up chart preview scenarios, chart data export, etc.
Think of a way to expose this functionality in another `igniteui-angular-integrations` repository, where we can add other functionalities like Summaries, Formatting, etc.

- [x] POC is done - https://www.infragistics.com/angular-demos/grid/grid-dynamic-chart-data
- [x] Directive implementation 
- [x] External repository creation?

HristoP96:
Closes #1667


src/app/grid/grid-dynamic-chart-data/directives/chart-integration/initializers.ts
```diff
@@ -74,28 +82,31 @@ export interface IStackedFragmentOptions {
     [key: string]: any;
 }
 
-export interface IChartComponentOptions {
-    chartOptions?: IChartOptions;
-    seriesOptions?: IChartSeriesOptions[];
-    xAxisOptions?: IXAxesOptions;
-    yAxisOptions?: IYAxesOptions;
-    stackedFragmentOptions?: IStackedFragmentOptions;
+export class ChartComponentOptions {
```
ddincheva: Are you sure that we need to switch this interface to class, according to the way you use it in you code?!

ddincheva:
When we select a range with keyboard and the Create Chart option is disabled.

PR #1671, opened 2020-01-21
fix(row-drag): fix row reorder `cursorPosition`

### LINKED ISSUE #1670: Issue with demo code not finding correct row index on drop after a row drag when page is scrolled.
## Description  
When a page is scrolled, and a grid is also scrolled. When we drag a row, on drop the demo code below finds the incorrect row index or in some cases depending on the page scroll, throws an error.

The issue is that getBoundingClientRect calculates the position as being relative to grid while pageX and pageY are relative to the page. The demo code is below, i have copied this code from this demo example -> https://stackblitz.com/angular/jlqrlpxkgdd. I have linked a stackblitz in attachments at the bottom of this page with the version of igx where it is breaking,

```typescript
    public onDropAllowed(args) {
        const event = args.originalEvent;
        const currRowIndex = this.getCurrentRowIndex(this.grid.rowList.toArray(),
            { x: event.pageX, y: event.pageY});
        // remove the row that was dragged and place it onto its new location
        this.grid.deleteRow(args.dragData.rowID);
        this.data.splice(currRowIndex, 0, args.dragData.rowData);
    }

    private getCurrentRowIndex(rowList, cursorPosition) {
        for (const row of rowList) {
          // relative to position inside the grid, therefore the scrolling the page can bring up the wrong position.
            const rowRect = row.nativeElement.getBoundingClientRect();
            if (cursorPosition.y > rowRect.top && cursorPosition.y < rowRect.bottom &&
                cursorPosition.x > rowRect.left && cursorPosition.x < rowRect.right) {
                // return the index of the targeted row
                return this.data.indexOf(this.data.find((r) => r.ID === row.rowData.ID));
            }
        }

        throw new Error("Failed to get row index.");
    }
```

 * igniteui-angular version:  8.2.15
 * browser: Version 79.0.3945.130 (Official Build) (64-bit)

## Steps to reproduce  

Step 1: Open the stackblitz linked below.
Step 2: Scroll the page to the bottom.
Step 3: Scroll the grid to the bottom.
Step 4: Pick up a row (prefer highest one that can be reached without scrolling)
Step 5: Drop the row on the bottom most row.
Step 6: Open console.

## Result  
Error in console saying failed to get row index. 

## Expected result  
On drop, we find the correct row index.

## Attachments  
I've got a stackblitz example here -> https://stackblitz.com/edit/angular-9z6azh

Error => 
image

ViktorSlavov:
Closes #1670 

Properly get the cursor coordinates in drop even


valadzhov:
Assigning the PR back to @ViktorSlavov. When you drag the first visible row a bit outside of the borders of the last row the "Failed to get row index" error is still being thrown.

valadzhov:
There is an error when you drag and drop a row outside the other rows in the Hierarchical Grid Row Reorder sample.
Hierachical Grid Row Reorder

ViktorSlavov:
@valadzhov , thanks for catching that!
it should now be addressed.

Issue #1681, opened 2020-01-22
Drag and Drop Samples not using Infragistics' Components


3phase:
### Description  
Related issue: #1631
The drag and drop samples are not using our components, which results in lack of/bad styling when switching from Light to Dark theme or when choosing a custom theme from the theming widget.

#### Affected samples ####
- [x] Dragging Using a Handle
- [x] Animations
- [x] Linking Drag to Drop Element (Email Sample)
- [x] Kanban Sample

### Steps to reproduce  

1. Open [Drag and Drop topic page](https://www.infragistics.com/products/ignite-ui-angular/angular/components/drag_drop.html)
2. Switch to Dark Theme
3. Scroll through the demo samples
4. Watch for bad styling, as for instance black font color on dark background or white drag handlers on white background

### Result  
Not using our components makes the elements not to get styled properly.

### Expected result  
If the native elements get replaced by our components, the respective element would get styled properly when switching from theme to theme.

PR #1682, opened 2020-01-23
Changes Email Sample List to igxList

3phase:
Related to issue #1681

PR #1683, opened 2020-01-23
Changes Email Sample List to igxList - master

3phase:
Related issue: #1681


MayaKirova:
I see a few minor issues:
1) Active list item style is lost on hover when dark theme is applied.
2) The drag item ghost is not styled according to the theme: 
image
3) Active style is not applied when item is selected in IE.

3phase:
> I see a few minor issues:
> 
> 1. Active list item style is lost on hover when dark theme is applied.
> 2. The drag item ghost is not styled according to the theme:
>    image
> 3. Active style is not applied when item is selected in IE.
> 4. Current background color is blue on both light and dark theme, perhaps it would be better to use the grid's selection style instead as it uses the secondary color:
>    background-color: var(--igx-grid-row-selected-background);

@MayaKirova, thank you for the remarks, I resolved them.

MayaKirova:
@3phase Could you also apply the changes to the vnext PR - https://github.com/IgniteUI/igniteui-angular-samples/pull/1682

3phase:
> @3phase Could you also apply the changes to the vnext PR - #1682

Done.

ChronosSF:
At this point it's too late to make master-specific fixes.

PR #1687, opened 2020-01-24
Refactor Drag and Drop animations sample

mddragnev:
Related issue: #1681


ChronosSF:
At this point it's too late to make master-specific fixes.

PR #1689, opened 2020-01-24
Refactor Drag dialog sample

mddragnev:
Related issue: #1681


MayaKirova:
Looks OK with the default themes. Once there are contrast colors exposed (https://github.com/IgniteUI/igniteui-angular/issues/6539)  we may be able to also change the text color to a contrasting one since the theme widget allows changing the surface color and it is possible to set it so the text is no longer visible, for example:
image

ChronosSF:
> Looks OK with the default themes. Once there are contrast colors exposed ([IgniteUI/igniteui-angular#6539](https://github.com/IgniteUI/igniteui-angular/issues/6539)) we may be able to also change the text color to a contrasting one since the theme widget allows changing the surface color and it is possible to set it so the text is no longer visible, for example:
> image

@MayaKirova , @3phase , we could wait for the contrast colors mixin but only if it is based on the surface color, which I think it is not.

3phase:
> > Looks OK with the default themes. Once there are contrast colors exposed ([IgniteUI/igniteui-angular#6539](https://github.com/IgniteUI/igniteui-angular/issues/6539)) we may be able to also change the text color to a contrasting one since the theme widget allows changing the surface color and it is possible to set it so the text is no longer visible, for example:
> > image
> 
> @MayaKirova , @3phase , we could wait for the contrast colors mixin but only if it is based on the surface color, which I think it is not.

The PR [IgniteUI/igniteui-angular#6539](https://github.com/IgniteUI/igniteui-angular/issues/6559) exposes a mixin, that generates a contrast-color property for all the colors and their shades, so it might resolve the unreadable text color in this pull request as well.

ChronosSF:
Awaiting the contrast color adjustments.

Issue #1720, opened 2020-02-03
Hierarchical grid samples using SINGERS data does not work properly


mddragnev:
### Description  
There is an invalid date in SINGERS data - release date.

### Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/paging.html
2. Expand the first row
3. Expand the album with name "Dream Driven"

### Result  
Observe that error is thrown

### Expected result  
No error should be thrown

### Attachments  
Attach a sample if available, and screenshots, if applicable.

PR #1722, opened 2020-02-03
fix(*): Make sure that date property is set properly

mddragnev:
Issue #1720


ChronosSF:
At this point it's too late to apply fixes to master

PR #1740, opened 2020-02-06
Move the host listener for changing themes to top level component 

dkamburov:
It seems that in some scenarios the listener is bound too late on Edge.


ChronosSF:
@dkamburov ,  would that fix the light/dark theme application that would still be used on IE (I assume this Edge bug also reproduces on IE)

MayaKirova:
> @dkamburov , would that fix the light/dark theme application that would still be used on IE (I assume this Edge bug also reproduces on IE)

@ChronosSF It seems that theme is reset to the default theme explicitly for IE in the docfx template:
https://github.com/IgniteUI/igniteui-docfx-template/blob/master/template/styles/theming-handler.js#L12

PR #1747, opened 2020-02-07
chore(*): fixing an error when the iframe tries to load multi font family from google apis

ChronosSF:
related to bootstrap default font-family


projects/app-lob/src/app/index/docs-layout.component.ts
```diff
@@ -26,7 +26,8 @@ export class DocsLayoutComponent implements OnInit {
             this.styleElem.textContent = e.data.themeStyle;
 
             const typeface = window.getComputedStyle(this.document.body).fontFamily.replace(/\"/g, "");
-            if (!this.typefacesLoaded.includes(typeface)) {
+            if (typeface.split(',').length < 2 &&
```
MayaKirova: Only 1 theme can be loaded in this way, so the checks should be <1.

PR #1782, opened 2020-02-14
chore(*): Adding logic to handle input type change if theme is changed.

### LINKED ISSUE #1757: Inputs do not comply with the applied theme
### Description  
When a new theme is applied from the theming service, the inputs are not changed based on the theme. For example, if you choose Bootstrap theme or Fluent theme, the inputs will stay Material, we should add `type` attribute to all input elements based on the selected theme `type="bootstrap"` and `type="fluent"`

<staging environment>/products/ignite-ui-angular/angular/components/input_group.html

MayaKirova:
Closes #1757 
Should be merged together with:
https://github.com/IgniteUI/igniteui-docfx-template/pull/164


ChronosSF:
@MayaKirova  This should be pr-ed to vNext as well

PR #1783, opened 2020-02-14
Fix toggle text visibility in the dark theme

### LINKED ISSUE #1733: In the dark themes the toggle text is not visible
### Description  
In the dark themes the toggle text is not visible. The same happens when you set custom themes via the theming widget.

### Steps to reproduce  
1. Open toggle samples
2. Select dark theme or change the theme via the theming widget
3. Open toggle

### Result  
Toggle text is not vissible

### Expected result  
Text should be vissible

### Attachments  
image

SisIvanova:
Closes #1733


simeonoff:
@SisIvanova This works correctly in the dark theme now, however it doesn't work with the theming widget. Please revise the toggle samples so that they work correctly with the theming widget.

Issue #1756, opened 2020-02-10
Dark themes make label tag unreadable.


mddragnev:
### Description  
Should be set some contrast color

### Steps to reproduce  

1. Open https://staging.infragistics.com/angularsite/components/grid/live_data.html
2. Apply material/fluent/bootstrap dark theme

### Result  
Observe that on the top of the grid labels Records and Frequency are black

### Expected result  
Labels should have contrast color so that be visible when dark theme is applied

### Attachments  
![image](https://user-images.githubusercontent.com/37667452/74148580-6ff80600-4c0e-11ea-8adc-3b8c2e8e2afd.png)

3phase:
The way this issue was resolved is via modifying the CSS of the sample itself. Even though, I think it would be better if the elements, styled via the typography properties, could change their font colors with respect to the selected theme, so that no custom CSS should be written when using one of our typography classes, for instance. 
@simeonoff, what do you think?

PR #1785, opened 2020-02-14
Fixes financial sample's slider labels' colors when a dark theme is applied - vNext

3phase:
Related issue: #1756


ChronosSF:
Think this should be pr-ed to master as well

PR #1804, opened 2020-02-20
Mkirkova/date range picker v next

### LINKED ISSUE #1739: add range-date-picker demo with calendar and input
create demo with calendar and input while range-date-picker is in development
#1574

MonikaKirkova:
closes #1739


valadzhov:
Currently are two issues with this sample:

- When you open and then just close the "From" dropdown menu an error is displayed in the console:

dateRange error

- When you select a city from the  "From" dropdown, close it and then expand it again you can't choose another city (this behviour is not exhibited in the "To" dropdown where you can change your choice:
FromTo

Lipata:
There were some comments from the Design Team, that needs to be addressed:

 - Maybe the range date picker should be on a new line (with a width equal to both autocompletes). This is needed because the calendar is cut in docfx:
image
- They wanted to remove the title in the sample, however, this is valid for all the available demos for date picker, that is why this could be done in a separate issue.
- Put some padding for the `Done` and `One Way Ticket` buttons

Other improvements:

- Either the autocomplete lists max height should be set or the height of the demo in docfx needs to be extended. Maybe the first one, because the demo should also look good in StackBlitz,
- The StackBlitz demo is opened but the inputs should be aligned better.

Issue #1817, opened 2020-02-24
Update column pinning sample


radomirchev:
Add column right pinning sample to the column pinning topic

PR #1825, opened 2020-02-25
Right Pinning Sample

3phase:
Related issues: #1817, IgniteUI/igniteui-angular#6676
_Test w/ latest master of igniteui-angular_


MayaKirova:
Stackblitz samples does not open:
image

PR #1827, opened 2020-02-27
Right Pinning Sample - Hierarchical Grid

3phase:
Related issues: #1817, IgniteUI/igniteui-angular#6676
PR for `igx-grid`: #1825
_Test w/ latest master of igniteui-angular_


MayaKirova:
Stackblitz sample does not load:
image

MayaKirova:
Also sample should allow pinning on children as well.

ChronosSF:
@MayaKirova , I added 3 pinned columns to children but I was not fan of having a toolbar as well as the child grids are very small and the toolbar would take most of the space.

Issue #1846, opened 2020-03-09
Refactor: Toast


StefanIvanov:
Improve the Toast samples.

PR #1858, opened 2020-03-16
refactor(toast): improve toast samples

SisIvanova:
Closes: #1846


src/app/notifications/toast/toast-style/toast-style.component.scss
```diff
@@ -1,31 +1,43 @@
-@import '~igniteui-angular/lib/core/styles/themes/index';
+.sample-content{
+    margin: 10% auto 0;
+}
+
+.sample-column {
+    min-width: 20vw;
```
simeonoff: This actually makes the sample less responsive. Take a look at the stackblitz sample and resize the preview to see how the buttons are not properly positioned.

src/app/notifications/toast/toast-sample-3/toast-sample-3.component.scss
```diff
@@ -1,5 +1,3 @@
-.div
-{
-    height: 100%; 
-    width: 100%;
-}
\ No newline at end of file
+button {
+    margin: 1% 1% 1% 0;
```
simeonoff: I'd use pixel values here, just so that samples are more visually consistent across devices.

src/app/notifications/toast/toast-style/toast-style.component.scss
SisIvanova: I changed the min-width to 270px (the default value is 300px) so that the three examples can be displayed in one line on smaller screens. Hope it looks better now.

PR #1883, opened 2020-03-23
Remote row pinning sample

dkamburov:


mddragnev:
@dkamburov I have found some issues regarding sorting and filtering in this sample.
The first one:
Steps:
1. Unpin the initially pinned rows and pin rows with id 1 and 2
2. Sort column Product Name for example.
3. Scroll down to the last record.
4. Observe the misalignment at the bottom of the grid.
Note: This is not reproducible in igniteui-angular dev samples 

Also:
Steps to reproduce:
1.Unpin both initially pinned records
2.Scroll to the middle of the data and pin any record. (for example with id 40)
3.Scroll to the top and pin record with id 1
4.Unpin the record with id 40.
5.Observe that above record is placed at the top of the unpinned area and after any interaction (sorting, scrolling, etc.) the record is placed where it was firstly.

Finally, as long as I know filtering should be applied to both pinned,unpinned collection separately.
Steps to reproduce:
1.Unpin both initially pinned records.
2.Filter column UnitPrice to equal 14.
3.Pin record with id 25 for example.
4.Remove the filtering value.
5.Pin the first record.
6.Filter the same column to equal 14.
7.Observe that in the pinned area the row with UnitPrice $18.00 is not filtered until any interaction with the grid( when you sort the column the filtering is applied too)

PR #1901, opened 2020-04-02
chore(*): Remove hideBaseOnDrag input as it is deprecated.

MayaKirova:
Related to https://github.com/IgniteUI/igniteui-angular/pull/7008


ChronosSF:
@MayaKirova , if you quickly grab a task and drag it outside the view and drop it, you can make it hide from the task list. It seems just hidden, but takes space and is counted in the task counters for the container.

PR #1911, opened 2020-04-09
Updating custom column pinning sample with correct pinning icon

### LINKED ISSUE #1898: Update the Custom Column Pinning UI sample with the correct pinning icon
Update the Custom Column Pinning UI sample to use the correct pin/unpin set of icons that are used in the excel-style filter.
Screenshot 2020-04-02 at 15 43 21

mddragnev:
Closes #1898


ChronosSF:
@mddragnev , stackblitz returns

Import error, can't find file:
../services/svgIcons

PR #1965, opened 2020-05-07
Keyboard nav hGrid sample

### LINKED ISSUE #1964: Keyboard navigation samples for hGrid and tGrid
### Description  
Keyboard navigation samples for hGrid and tGrid

Aleksandyr:
Closes #1964


nrobakova:
Hi @Aleksandyr ,
I found some issues in this sample:
1. When press Alt+L - this is not checked in the grid
2. When press Ctrl+Shift+l this is not checked in the list
3. When open a cell in edit mode and press Tab the list looks like the grid is not focused

Aleksandyr:
@nrobakova @zdrawku The first two issues are not related to the sample. That bizarre behavior is reproducible only when we try to execute those combinations under STG environment. I have tried to reproduce it both locally and by opening the sample directly on stackblitz and it was working perfectly.

The last issue was resolved with the last push.

HristoP96:
> @nrobakova @zdrawku The first two issues are not related to the sample. That bizarre behavior is reproducible only when we try to execute those combinations under STG environment. I have tried to reproduce it both locally and by opening the sample directly on stackblitz and it was working perfectly.
> 
> @HristoP96 Any suggestions?
> 
> The last issue was resolved with the last push.

The sample on staging and the sample on stackblitz are different, here's why:
- The checkboxes are different:
image
- The angular version is different:
image
 So I assume the staging build might have changed some things, but this needs some investigation

Aleksandyr:
@HristoP96 @zdrawku so should we redeploy after approving this PR?

zdrawku:
Yes, we should

PR #1976, opened 2020-05-11
docs(dock-manager): Create dock manager sample

onlyexeption:


src/app/layouts/dock-manager/dock-manager.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="sample-container">
```
DiyanDimitrov, line 1: Remove the sample-container div.

package.json
```diff
@@ -40,6 +40,7 @@
     "@angular/platform-browser": "9.1.0",
     "@angular/platform-browser-dynamic": "9.1.0",
     "@angular/router": "9.1.0",
+    "@infragistics/igniteui-dockmanager": "file:infragistics-igniteui-dockmanager-0.1.0-beta.5.tgz",
```
DiyanDimitrov, line 43: Use the proget package and delete the local one.

src/app/layouts/dock-manager/dock-manager.component.html
DiyanDimitrov: Remove the whole div, not only the class.

PR #1977, opened 2020-05-12
Adding text selection directive samples

### LINKED ISSUE #1966: Update the igxMask samples with igxTextSelection directive
### Description  
The samples showing the `igxMask` directive should include the `igxTextSelection` directive. As a user, I would like to see the mask selected when I'm focusing the input that contains it, so I can start typing directly.

HristoP96:
Closes #1966


zdrawku:
For all modified samples both the sample app and Stackblitz config are working properly

PR #1988, opened 2020-05-13
Fixing IE 11 rendering issue

### LINKED ISSUE #1983: In IE11 samples does not have styles applied
### Description  
In IE11 samples does not have styles applied

### Steps to reproduce  

1. Open [sample](https://www.infragistics.com//angularsite/components/grid/editing.html) or othe sample in IE 11

### Result  
The grid styling is not applied and there is an error in the console

### Expected result  
Grid should look correct as in other browsers

### Attachments  

image

HristoP96:
Closes #1983


HristoP96:
@simeonoff  This Pr fixes the rendering problem in 11, however it enables the $legacy-support of the theming framework for both **app** and **app-lob** applications. Should  we think of an approach for disabling the **$legacy-support**,  when the application is run with **esnext** configuration?

simeonoff:
> @simeonoff This PR fixes the rendering problem in 11 by enabling the $legacy-support of the theming framework for both **app** and **app-lob** applications. Should we think of an approach for disabling the **$legacy-support**, when the application is run with [**esnext** ](https://github.com/IgniteUI/igniteui-angular-samples/blob/master/package.json#L14) configuration?

That would be ideal, but let's not block this PR.

PR #1998, opened 2020-05-15
Enter key in Custom keyboard navigation samples

nrobakova:
Closes ##1995


Aleksandyr:
@nrobakova It seems that there is some problem in the tGrid and Grid samples.
The editing is executing no more than twice. If I try to proceed with more then 2 operations nothing happens.

nrobakova:
@Aleksandyr  this happens because of this [issue ](https://github.com/IgniteUI/igniteui-angular/issues/7345) in the product. It is fixed and when the new version is ready the sample will be ok :)

PR #2011, opened 2020-05-19
fix(slider): two-way databound slider inputs

### LINKED ISSUE #2000: Input issues with the Slider sample
## Description  
https://www.infragistics.com/products/ignite-ui-angular/angular/components/slider.html
1. The slider thumb does not change its position when the user types value in the inputs and confirm it by ENTER
2. The slider thumb does not change its position when the user edits ONLY part of the value, without editing the $ dollar sign, e.g. from $200 to $300.

 * igniteui-angular version: 8.2.12
 * browser: all

## Steps to reproduce  

1. Change the value in the first input for the idle value $200 to $300, only by changing the digits;
2. Press Enter

Note that if the whole value, including the $ dollar sign is changed, then TAB button triggers the thumb to change its position. But ENTER does not. Moreover, if I, as an end user, change only the digits, without the $ dollar sign, then even the TAB button does not trigger the thumb.

## Result  
The first thumb position does not change its position.

## Expected result  
No matter changing the whole value or just the digits, after ENTER the new input data should be confirmed and lead to change in the position of the sliders thumb

Aleksandyr:
Closes #2000


hanastasov:
Start typing 100 in the UpperInput field. On typing the second `0`, value is converted autoamtically to 800. I think this needs to be fixed.

Aleksandyr:
@hanastasov This is the default behavior of the slider as it validates the values of the lower and upper thumbs in a way that keeps their boundaries correct. The lower could never exceed the upper and vice versa.

PR #2012, opened 2020-05-19
Add dock manager config generator

onlyexeption:


DiyanDimitrov:
@onlyexeption, Could you please replace the private licensed package with the [npmjs one](https://www.npmjs.com/package/igniteui-dockmanager) and check if the stackblitz sample works as expected?

HristoP96:
@onlyexeption  @DiyanDimitrov  I added the dock manager to the additional dependencies of the config, however, the Stackblitz bundle is not loading the package after the the bundles instantiation, this may be due the beta version of the package

HristoP96:
@DiyanDimitrov @onlyexeption There must be applied additional logic in the live-editing generator, due the additional requirements that dock manager needs in order to work properly, like define elements in the main.ts and adding schemas property in the ngModule in app module.ts.

HristoP96:
I will let you know when I am ready with the live editing update

Issue #2006, opened 2020-05-18
Update remote paging samples/Add new sample for `pagingMode`


zdrawku:
### Description  
Based on this [feature request](https://github.com/IgniteUI/igniteui-angular/issues/7215#issuecomment-630146450), create a new sample for Remote Paging with Batch editing and update the old ones.

PR #2022, opened 2020-05-21
Remote paging with batch editing

ddincheva:
Related to #2006


ddincheva:
@hanastasov , in the provided example we use only public API, so it is absolutely valid solution for the described use case scenario. When you have local and remote data this is the only way for managing them. If the user wants to mix them, he has to either handle all operation remotely or to provide all the data to the grid/locally/ and left to the grid to handle these cases.

hanastasov:
> @hanastasov , in the provided example we use only public API, so it is absolutely valid solution for the described use case scenario. When you have local and remote data this is the only way for managing them. If the user wants to mix them, he has to either handle all operation remotely or to provide all the data to the grid/locally/ and left to the grid to handle these cases.

Fair enough :)

Issue #1991, opened 2020-05-14
Issue w/ Redo-button in the Grid Paste from Excel Sample


3phase:
### Description  
When clicking on the Redo button of the /grid/grid-paste sample, the grid won't scroll to the bottom, as expected.

### Steps to reproduce  

1. Open the /grid/grid-paste sample;
2. Paste several records from Excel;
3. Click the Undo button to remove the last one;
4. Click the Redo button to restore the removed record;

### Result  
The record gets restored, but the scroll position remains the same, instead of scrolling to the bottom of the grid.

### Expected result  
Scroll position to change in order the last record to get shown.

### Why does it happen?
The `scrollTo` gets invoked prior the appending of the restored record to the `verticalScrollContainer.igxForOf` collection. If one places the `scrollTo` code snippet inside a `setTimeout`, for instance, the issue won't get reproduced, because the record would have already been added to the collection.

PR #2039, opened 2020-05-27
Add detectChanges when redo changes in grid paste sample

mddragnev:
Related issue: #1991


mddragnev:
Hello @damyanpetev,
In my opinion the issue #1991 is caused because the igxForOf does not detect if there is any change made by transaction service **undo** . Do you think that it should be somehow fixed in there or leave it on sample level?

ChronosSF:
@damyanpetev , what do you think?

damyanpetev:
@mddragnev @ChronosSF Sorry for the delay.
Okay, so the issue is with the sample as far as I can tell - basic reason being modifications to the data take a cycle for the pipes to kick up and update the `igxFor` state. Can't add a comment on that line, but look at the `addRecords` method that apparently had the same issue and solved it the same way - pumping the `detectChanges` button in the for loop no less 😐

Forcing change detection is kinda "big hammer for a small problem" (especially repeatedly) and could be costly or larger, more complicated grids. Also, keep in mind, `addRow` will retrigger on its own and the undo/redo emit [`onStateUpdate`](https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/igxtransactionservice.html#onstateupdate) that the grid is subscribed to and that will also do the same.

So what you are looking for is a way to scroll _when_ the data is rendered, so how about `onDataChanged`? Bonus, since you're actually waiting on the data view to refresh for the scroll, you might as well use it for the index instead of calculating it with transactions:
```ts
// scroll to last row
this.grid1.verticalScrollContainer.onDataChanged.pipe(first()).subscribe(() => {
  this.grid1.verticalScrollContainer.scrollTo(this.grid1.dataView.length - 1);
});
```
This works both after add or undo: https://stackblitz.com/edit/grid-paste-redo?file=src%2Fapp%2Fgrid%2Fgrid-paste-sample.component.ts 

PS: While at it, there seems to be an unused `addedData` array and the 4 space formatting looks rather unfortunate in Stackblitz as double indent 🙃

damyanpetev:
PPS: Didn't the grid gain some very public `navigateTo` API? I suppose that can be used instead as well

src/app/grid/grid-paste/grid-paste-sample.component.ts
```diff
@@ -16,7 +16,7 @@ import { PasteHandler} from "./paste-handler.directive";
 
 import { EXCEL_DATA, LOCAL_DATA } from "./data";
 
-import { take } from "rxjs/operators";
+import { first, take } from "rxjs/operators";
 
 @Component({
     encapsulation: ViewEncapsulation.None,
```
damyanpetev, line 22: Since I can comment on this now - no idea what the sample is doing without encapsulation, but it shouldn't. This should be removed

mddragnev:
Hello @damyanpetev , I think I did the things you mentioned above. Please let me know if there is something else that bothers you.

PS: I don`t have a template validation error when running the sample locally.

skrustev:
- [ ] For the paste option "starting from active cell" when performing the same steps it should not scroll to the bottom but it should either retain the scroll position.

mddragnev:
@skrustev Should be fixed now.

src/app/grid/grid-paste/grid-paste-sample.component.html
```diff
@@ -1,8 +1,8 @@
 <div class="grid__wrapper">
     <div class="grid__menu">
-        <button igxButton="raised" (click)="toggleDropDown($event)">Paste Options</button>
-        <igx-drop-down (onSelection)="selectionChange($event)">
-            <igx-drop-down-item *ngFor="let item of comboData">
+        <button igxButton="raised" [igxToggleAction]="dropdown" [igxDropDownItemNavigation]="dropdown">Paste Options</button>
+        <igx-drop-down #dropdown (onClosing)="changePasteMode()">
```
damyanpetev, line 4: fairly certain there's selectionChanged event you can use and skip the viewchild :)

src/app/grid/grid-paste/grid-paste-sample.component.scss
```diff
@@ -25,14 +13,6 @@
...
 
-.header-icon {
-    font-size: 1.4em;
-    width: 1.1em;
-    height: 1.1em;
-    float: right;
-    cursor: pointer;
-}
-
 .header {
```
damyanpetev, line 16: there's no `header` class applied anywhere in the template, so this likely doesn't do anything. Same for the `igx-grid__th` styles I think

src/app/grid/grid-paste/grid-paste-sample.component.ts
```diff
@@ -126,14 +126,17 @@ export class GridPasteSampleComponent {
         for (const pkVal of updatedRecsPK) {
             const row = this.grid1.getRowByKey(pkVal);
             if (row) {
-            row.nativeElement.style["font-style"] = "italic";
-            row.nativeElement.style.color = "gray";
+                row.nativeElement.style["font-style"] = "italic";
```
damyanpetev, line 129: Huh, just noticed this bit... I don't get it, the transactions integration with the grid already applies "modified" styles, this shouldn't be needed at all. Especially, since modifying the row native elements is usually not a good idea with virtualization. Can you check if this is needed and remove it if not?

PR #2079, opened 2020-06-09
Update custom styles in some samples #1418

### LINKED ISSUE #1418: Primary and secondary colors are not the same in the different styling samples
### Description  
Looking at the different "Styling" sections in the docs, the colors used in the samples are different.
We should use the same colors to achieve consistency between the Styling samples.
Here are the colors that some samples are using:
$yellow-color: #FFCD0F;
$black-color: #292826;

### Steps to reproduce  
1. The samples that should be changed:
https://www.infragistics.com/angular-demos/grid/grid-column-hiding-style (should use the colors above)
https://www.infragistics.com/angular-demos/grid/custom-grid-paging-style (page size drop down uses pink color)
https://www.infragistics.com/angular-demos/grid/grid-filtering-style (condition drop down uses default pink color)

### Result
Inconsistent colors in the Styling samples

### Expected result  
Have consistent colors between different Styling samples

VMihalkov:
Closes #1418


mddragnev:
Hello @VMihalkov @ChronosSF ,
In custom grid paging style sample when items per page drop down is expanded, the number of items is still selected with pink color.

PR #2086, opened 2020-06-11
fix(select): adjust select item content display to account for lib change

ViktorSlavov:
Depends on https://github.com/IgniteUI/igniteui-angular/pull/7329

With the change to the content of the `igx-drop-down__item` (and, therefore, the select item), we need to make some changes to our sample in order to preserve their look:

The content of the select item is no longer `display: flex`, so we need to manually add that.
Since the item's content is part of the `igx-drop-down-item` template, we need to penetrate view encapsulation in both samples
To prevent style from "leaking", we need to scope by ':host' and use a display `outlet` for the overlay.
Finally, one of the samples was using `ViewEncapsulation.None` and was leaking styles - this is addressed.

### DIFF
--- a/src/app/data-entries/select/select-header-footer/select-header-footer.component.scss
+++ b/src/app/data-entries/select/select-header-footer/select-header-footer.component.scss
@@ -9,8 +9,15 @@
     margin-left: 48px;
 }
 
-.igx-drop-down__item {
-    justify-content: space-between;
+:host ::ng-deep {
+    .igx-drop-down__item {
+        .igx-drop-down__inner {
+            display: flex;
+            padding: 8px;
+            justify-content: space-between;
+            width: 100%;
+        }
+    }
 }
 
 igx-select {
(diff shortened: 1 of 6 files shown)


src/app/data-entries/select/select-header-footer/select-header-footer.component.scss
```diff
@@ -9,8 +9,15 @@
     margin-left: 48px;
 }
 
-.igx-drop-down__item {
-    justify-content: space-between;
+:host ::ng-deep {
+    .igx-drop-down__item {
```
simeonoff, line 13: Do not target inner classes. Add "user-specific" classes directly to the element or use the element selector instead. We are trying to teach people how to do things the right way. They shouldn't rely on the component classes we use, as they can change.

src/app/data-entries/select/select-sample-2/select-sample-2.component.scss
```diff
@@ -8,3 +8,14 @@
     padding-left: 10px;
     padding-left: 10px;
 }
+
+:host ::ng-deep {
+    .igx-drop-down__item {
```
simeonoff, line 13: Same as above.

PR #2089, opened 2020-06-15
fix(select): adjust select item content display to account for lib change - vNext

ViktorSlavov:
Depends on https://github.com/IgniteUI/igniteui-angular/pull/7329

With the change to the content of the `igx-drop-down__item` (and, therefore, the select item), we need to make some changes to our sample in order to preserve their look:

The content of the select item is no longer `display: flex`, so we need to manually add that.
Since the item's content is part of the `igx-drop-down-item` template, we need to penetrate view encapsulation in both samples
To prevent style from "leaking", we need to scope by ':host' and use a display `outlet` for the overlay.
Finally, one of the samples was using `ViewEncapsulation.None` and was leaking styles - this is addressed.


ViktorSlavov:
This is only needed in `vNext` branch

PR #2100, opened 2020-06-25
Update grid edit events sample

wnvko:
Make the sample to use the new properties exposed in `onCellEdit` event arguments 

Related: https://github.com/IgniteUI/igniteui-docfx/issues/2039


Lipata:
Cherry-picker into another PR: https://github.com/IgniteUI/igniteui-angular-samples/pull/2111

PR #2102, opened 2020-06-26
Promote v10

zdrawku:


zdrawku:
let's not merge it yet, we are investigating a problem on staging environment that might occur on production as well if we merge this PR

PR #2163, opened 2020-07-31
Update the overlay styling sample

desig9stein:
Update the sample to show how to style the overlay component.


desig9stein:
> Unfortunately, this PR doesn't fix the issue.
> image

This is one part of the fix, to make the sample work like any other styling sample, you need to have [this](https://github.com/IgniteUI/igniteui-docfx/pull/2132/files) PR merged.

PR #2167, opened 2020-08-04
chore: ng update igniteui-angular@10.1.0-alpha.5

Lipata:


Lipata:
It's still not necessary, but should be updated at some point, either by migration or manually. @zdrawku, @ddincheva can you do that in a separate PR?

ddincheva:
@Lipata , Yes, I will address the necessary changes in a separate PR :)

Issue #2096, opened 2020-06-24
Splitter sample is loaded with scrollbars that are hiding the *handlebar*


zdrawku:
### Description  
I believe this sample can be improved by removing some of the *Lorem ipsum* content. When the sample is loaded you see two gray scrollbars and the *splitter delimiter/handlebar* is lost:

![image](https://user-images.githubusercontent.com/11193764/85520355-77430d80-b60b-11ea-9cae-7c3419d4d414.png)

Also, why some splitter panes are having scrollbars with a smaller width than others? Shouldn't they be with a consistent width?

![image](https://user-images.githubusercontent.com/11193764/85520479-a8bbd900-b60b-11ea-931c-0e5a972a1a37.png)

PR #2196, opened 2020-08-12
Change text length and splitter pane size of splitter samples

mddragnev:
Related issue: #2096


VMihalkov:
I think that more _Lorem ipsum_ content needs to be removed, as the scrollbars are still visible in the first sample. My understanding from the issue is that there should be no scrollbars. Here is how the sample looks like now. There is not much difference from before the fix:
image

mddragnev:
@VMihalkov Should be good now

Issue #2088, opened 2020-06-11
In sample for remote data operation when add 2 filtering conditions an error is returned


nrobakova:
### Description  
In sample for remote data operation when add 2 filtering conditions an error is returned

### Steps to reproduce  

1. Open [sample ](https://www.infragistics.com//angularsite/components/grid/remote_data_operations.html#grid-remote-data-operations-overview)
2. Open filter row for a column and add 2 filter conditions

### Result  
An error in console is returned and the grid data is not updated

### Expected result  
Grid should be filtered and display filtered data

### Attachments  
![image](https://user-images.githubusercontent.com/11463104/84402437-167c0400-ac0d-11ea-8ab9-524cf4113574.png)

PR #2200, opened 2020-08-13
Parsing the filtering expression tree when having multiple conditional filtering in remote filtering sample

mddragnev:
Related issue: #2088


src/app/grid/grid-remote-filtering-sample/remote-filtering-sample.component.ts
```diff
@@ -76,7 +76,10 @@ export class RemoteFilteringSampleComponent implements OnInit {
         }
 
         const virtualizationState = this.grid.virtualizationState;
-        const filteringExpr = this.grid.filteringExpressionsTree.filteringOperands;
+        const filteringExpr = [];
+        this.grid.filteringExpressionsTree.filteringOperands.forEach(operand => {
+            filteringExpr.push(this._remoteService.getFilteringExpressionTreeForColumn(operand))
+        });
```
DiyanDimitrov, line 82: The approach is fine, but I think it would be better to make this transformation in the remote service's `_buildAdvancedFilterExpression` method.

DiyanDimitrov (approved):

Issue #2005, opened 2020-05-18
Refactor the Virtual For-Of Samples


simeonoff:
Improve the virtual for-of samples and documentation.

PR #2205, opened 2020-08-18
refactor(igxFor): improve the virtual for-of samples

SisIvanova:
Closes: #2005


simeonoff (changes requested):
The horizontal and vertical samples can be improved more. The horizontal sample is not working correctly at the moment as its items resize when the page is resized. All items should have stable size, regardless of the width of the window.

PR #2241, opened 2020-09-11
Refactor Live Editing Generator

### LINKED ISSUE #2187: Prevent  the transpilation of web libraries during live editing generation
### Description  
Currently the transpilation of the LiveEditingManager includes every library, used in the samples as dependency. We have to think of a way to prevent this, because the live editing generation is not within the browser context.

HristoP96:
Closes #2187


zdrawku (changes requested):
The application is executed correctly as well as loading within the iframes, although there is some kind of a problem with the live-editing with the app-lob samples, the component files are not in the right app location. For example:
image

image

And StackBlitz:
image


This might be related to this method, I will need to debug it for more information:
https://github.com/IgniteUI/igniteui-angular-samples/pull/2241/files?file-filters%5B%5D=.ts&hide-deleted-files=true#diff-595fd4713046b0c4ccff27dc70b7cf6aR157

zdrawku (changes requested):
SB and CodeSandbox samples are still not loading:
image

As you pointed out, there is a problem with the path backslash escaping based on the dev environment (Windows or Linux)

zdrawku (changes requested):
Slash direction and path are not correct
image

Remove projects/app-lob from the path

HristoP96:
> Slash direction and path are not correct
> image
> 
> Remove projects/app-lob from the path

The fix is now available.

zdrawku (approved):

PR #2273, opened 2020-10-06
Add tree grid add row samples

### LINKED ISSUE #2269: Create Add Row Tree Grid Samples
- [ ] Sample deriving from the Row Editing sample showing default native row adding UI for tree grid without transactions.
    - specifically show conditional "add child" visibility in the action strip template (such as hide it if the row's level is more than 2)
- [ ] Sample deriving from the first one showing custom styling of all UI elements
    - the styling of the edit UI can be taken from the Row Editing styling sample
    - some styling can be applied to the icons in the action strip
    - some other styling can be applied to the unfiled values (column headers) text

VMihalkov:
Closes #2269 

Added 2 samples for tree grid with add row.


ChronosSF (changes requested):
these seem to lack the live-editing data

VMihalkov:
> these seem to lack the live-editing data

Configs for live-editing have been added for the new samples

PR #2275, opened 2020-10-06
Add flat grid add row samples

### LINKED ISSUE #2262: Create Add Row Flat Grid Samples
- [ ] Sample deriving from the Row Editing sample showing default native row adding UI for flat grid without transactions
- [ ] Sample deriving from the first one showing custom styling of all UI elements 
    - the styling of the edit UI can be taken from the Row Editing styling sample
    - some styling can be applied to the icons in the action strip
    - some other styling can be applied to the unfiled values (column headers) text

teodosiah:
Closes #2262 

Added 2 samples for a flat grid with add row functionality.


ChronosSF:
Both samples load correctly in codesandbox and stacblitz and also function correctly.

@simeonoff , could you please check the styling on the style-related sample as I think we can make it prettier. I think it uses the theme from the row editing sample but that one is simpler and the colors don't translate well on this one. I also think we wanted to use the black and yellow theme for styling samples across the board. Up to you really.

simeonoff (changes requested):
I would remove the styling sample altogether. In the DocFX article, we can simply explain which components comprise the feature and link to their respective styling sections.

ChronosSF (approved):

PR #2289, opened 2020-10-14
Create preset overlay settings sample

PlamenaMiteva:


src/app/interactions/interactions-routing.module.ts
```diff
@@ -247,6 +248,11 @@ export const interactionsRoutes: Routes = [
         data: interactionsRoutesData["overlay-scroll-sample-2"],
         path: "overlay-scroll-sample-2"
     },
+    {
+        component: OverlayPresetSettingsSampleComponent,
+        data: interactionsRoutesData["overlay-relative-presets-sample"],
```
wnvko, line 253: the new sample was added with key `overlay-preset-settings-sample` in `interactionsRoutesData`. The key here is wrong.
```suggestion
        data: interactionsRoutesData["overlay-preset-settings-sample"],
```

wnvko (changes requested):
The new sample should be added to `OverlayConfigGenerator.ts`.

wnvko (changes requested):
The new sample should be added to `OverlayConfigGenerator.ts`.

Lipata:
@PlamenaMiteva, please create this pull request only from vNext branch.

PR #2295, opened 2020-10-15
fix(tsconfig): change target to es5

### LINKED ISSUE #2258: CRM grid Error: strict mime type checking is enforced for module scripts per html spec
### Description  
CRM grid is not loaded under Edge
image

zdrawku:
Closes #2258


HristoP96 (approved):

HristoP96 (changes requested):
Shouldn't this change be applied in this [file](https://github.com/IgniteUI/igniteui-angular-samples/blob/master/projects/app-lob/tsconfig.app.json)

zdrawku:
Closing this PR, the problem is not because of the ES compilation

Issue #2310, opened 2020-10-19
Hierarchical grid add row sample breaks if we add row from a record with duplicate PK value


mddragnev:
### Description
We should probably change the data so that no PK value is duplicated

### Steps to reproduce  

1. Open hierarchical grid add row sample
2. Scroll down to a record with name
`Pablo Cambeiro`
3. Add row from it

### Result  
Observe that the row is trying to be added from the first row in the grid

### Expected result 
The row should be added from the  `Pablo Cambeiro` row.

PR #2313, opened 2020-10-20
Add ID field of SINGERS data so that there is no duplication in PK values

mddragnev:
Closes: #2310


ChronosSF (approved):

ChronosSF:
@mddragnev , this should be pr-ed to master as well

PR #2402, opened 2020-12-04
Update FinJS demo - vNext

hanastasov:


dobromirts (changes requested):
When "Live all prices" is triggered, it does not change the value of the price, instead it only changes the material-icon and the color.

When the page is initialized and the "Chart" button is pressed, the chart is not displayed.
emptyChart

dobromirts (approved):

zdrawku (approved):

PR #2404, opened 2020-12-07
refactor(drop-down): include prefix, suffix and divider sample

SisIvanova:
Related to: IgniteUI/igniteui-angular/issues/8656

### DIFF
--- a/src/app/data-entries/dropdown/dropdown-sample-3/dropdown-sample-3.component.scss
+++ b/src/app/data-entries/dropdown/dropdown-sample-3/dropdown-sample-3.component.scss
@@ -1,8 +1,12 @@
 .button {
     margin: 8px;
-    width: 160px;
+    width: 240px;
 }
 
 .drop-down__scroll-container {
     max-height: 240px;
+}
+
+.igx-drop-down__header igx-icon{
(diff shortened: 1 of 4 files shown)


simeonoff (approved):

simeonoff (changes requested):
Update the DropDown config generators so that the sample works in StackBlitz and CodeSandBox.

src/app/data-entries/dropdown/dropdown-sample-3/dropdown-sample-3.component.scss
```diff
@@ -1,8 +1,12 @@
...
     margin: 8px;
-    width: 160px;
+    width: 240px;
 }
 
 .drop-down__scroll-container {
     max-height: 240px;
 }
+
+.igx-drop-down__header igx-icon{
```
simeonoff, line 10: We should also not target the inner class selector directly. Try using a different selector, for instance:

```scss
igx-drop-down-item[isHeader="true"] igx-icon {
...
}
```

simeonoff (changes requested):

simeonoff (approved):

PR #2415, opened 2021-01-04
Formatting  app.module.ts live editing output file 

### LINKED ISSUE #2414: Remove indentation in module files of code view
Currently the `Code View` shows all `imports` and `declarations` that we include during generation with additional indentation.

image

While this makes it a little bit easier for a person to differentiate which items were added by us, we could consider removing that indentation as it also looks kind of broken.

HristoP96:
Closes #2414

### DIFF
--- a/live-editing/generators/SampleAssetsGenerator.ts
+++ b/live-editing/generators/SampleAssetsGenerator.ts
@@ -184,37 +184,39 @@ export class SampleAssetsGenerator extends Generator {
     }
 
     private _getAppModuleConfig(config: Config, configImports, configAdditionalImports?) {
+        let defaultNgDeclarations = ["AppComponent"];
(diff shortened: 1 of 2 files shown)


HristoP96:
@jackofdiamond5 This PR will fix the indentation problem, which persists in the code view, here is an example on local run:
image

FYI, there will be some indentation inconsistency between the code view and the codesandbox/Stackblitz bundles (here is the codesandbox of this sample):
image
However the CSB/Stackblitz editors have  integrated formatters, so I think this inconsistency is not a big deal :)

jackofdiamond5 (approved):

dobromirts (approved):

zdrawku (approved):
Nice, you are concatenating the declarations and imports thus the removal of the hardcoded declarations and imports.
@HristoP96 could you please cherry-pick these changes to master

image

PR #2421, opened 2021-01-07
fix(theming): missing components roundness sample

SisIvanova:


simeonoff (approved):

src/app/data-entries/buttons/buttons-roundness-sample/buttons-roundness-sample.html
```diff
@@ -0,0 +1,26 @@
+<div class="sample-wrapper">
```
simeonoff, line 1: You don't need this "sample-wrapper" container.

simeonoff (changes requested):

simeonoff (approved):

PR #2443, opened 2021-01-21
Refactor samples with new toolbar -vnext

### LINKED ISSUE #2431: `columnPinning`, `toolbarTitle` are deprecated
Some samples are not migrated to the new toolbar.
Example [StackBlitz](https://stackblitz.com/github/IgniteUI/igniteui-live-editing-samples/tree/master/angular-demos/grid/grid-sample-toolbar-pinning) part of https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-pinning

dobromirts:
Closes #2431

### DIFF
--- a/projects/app-lob/src/app/grid-finjs/grid-finjs.component.html
+++ b/projects/app-lob/src/app/grid-finjs/grid-finjs.component.html
@@ -3,23 +3,24 @@
         igxPreventDocumentScroll
         width="100%" height="30%"
         displayDensity='compact'
-        columnHidingTitle="Indicators"
         hiddenColumnsText="Hidden"
         primaryKey='ID'
-        exportExcelText="Export to Excel"
         [autoGenerate]='false'
         [data]="data"
         [rowSelection]="selectionMode"
         [allowFiltering]="true"
         [filterMode]="'excelStyleFilter'"
-        [columnHiding]="true"
-        [columnPinning]="true"
-        [exportExcel]="true"
         (onRowSelectionChange)="rowSelectionChanged($event)"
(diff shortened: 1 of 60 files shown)


projects/app-lob/src/app/grid-finjs/grid-finjs.component.html
```diff
@@ -3,23 +3,24 @@
         igxPreventDocumentScroll
         width="100%" height="30%"
         displayDensity='compact'
-        columnHidingTitle="Indicators"
         hiddenColumnsText="Hidden"
         primaryKey='ID'
-        exportExcelText="Export to Excel"
```
zdrawku, line 9: The ability to set excel export text and csv export text should be exposed as well. Example:

```
<igx-grid-toolbar-exporter [exportCSV]="false" [exportExcel]="true">
        <span excelText>Transform to Excel</span>
        <span csvText>Transform to CSV</span>
</igx-grid-toolbar-exporter>
```

I assume the default values of exportCSV and exportExcel equivalents was 'false' and thats the reason to explicitly set it?

Please do consider this for all following examples below.

zdrawku (changes requested):

projects/app-lob/src/app/grid-finjs/grid-finjs.component.html
zdrawku: Also, we should check why there is no equivalent for "hiddenColumnsText" which will be deprecated. We might have be facing a bug here ;)

zdrawku (approved):

PR #2484, opened 2021-02-22
Add workaround for chart area in finjs sample

HristoP96:


projects/app-lob/src/app/grid-finjs/main.component.scss
```diff
@@ -5,6 +5,10 @@
     }
 }
 
+.chart-container{
```
zdrawku, line 8:
```suggestion
.chart-container {

```

zdrawku (approved):

PR #2493, opened 2021-02-26
Expose RTL support sample -vnext

dobromirts:
Related to #https://github.com/IgniteUI/igniteui-docfx/issues/2562


zdrawku (changes requested):
Code view element and live-editing buttons are not visible:
image

Also, please remove the localization change logic (the drop down) and move the button "Change directionality" at the top.

zdrawku (approved):

PR #2503, opened 2021-03-08
refactor(*): use-elevations-as-css-vars

SisIvanova:
Related to: IgniteUI/igniteui-angular/issues/8920


src/styles.scss
```diff
@@ -5,7 +5,7 @@
 @include igx-theme($default-palette, $legacy-support: true);
 
 :root {
-    @include css-vars-from-palette($default-palette);
+    @include igx-palette-vars($default-palette);
```
simeonoff, line 8: This is no longer needed as the palette colors are included as CSS variables by default now.

simeonoff:
vNext should work fine without the changes in this PR.

PR #2517, opened 2021-03-15
Fix live editing for Live-data grid - master

### LINKED ISSUE #2515: Live editing of Live-data grid is not working
1. Click on any of the live-editing buttons

dobromirts:
Closes #2515


HristoP96 (changes requested):
The stackblitz bundle complains about the resize-observer-polyfill missing from the dependcies, because it is used in the sample business logic.

@dobromirts  Can add you the this package to the [additionalDependecies](https://github.com/IgniteUI/igniteui-live-editing/blob/master/src/public.ts#L45) property of the finjs config

HristoP96 (approved):

zdrawku (approved):

PR #2543, opened 2021-03-24
FinJS sample with DockManager enhancements --master

### LINKED ISSUE #2541: FinJS sample with DockManager enhancements
Let's address the following points:
- [x] The service should perform 30 updates per frequency value (300, 600, 900) not 10 000.
- [ ] The ability to change frequency and volume should remain only for the Main Grid (the grid with initial grouping)
- [ ] All other grids should have 100 records (that includes the dynamically created grids as well)
- [x] Initially load the Main Grid with 500 records
- [x] Add select item with value 100 (currently have 500, 1000, 5000, 10 000)
- [x] There is a filtering overlay bug, and when the filtering is opened it will be hidden by the other dock-manager slots. It should appear always on top.  This comes from the grid itself - use [this suggestion](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/excel-style-filtering#external-outlet) in order to overcome the problem.
- [x] Select elements are always with a white theme.

ddincheva:
Closes #2541

### DIFF
--- a/projects/app-lob/src/app/grid-finjs-dock-manager/grid-finjs-dock-manager.component.html
+++ b/projects/app-lob/src/app/grid-finjs-dock-manager/grid-finjs-dock-manager.component.html
@@ -9,9 +9,8 @@
             </div>
             <div class="actionItem">
                 <!-- Change volume -->
-                <igx-select [(ngModel)]="dataVolume" (ngModelChange)="paramsChanged()">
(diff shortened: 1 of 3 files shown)


kdinev:
@ddincheva Merging this so we can get it on production ASAP. Will leave the `awaiting-test` label.

kdinev (approved):

PR #2550, opened 2021-03-29
Fix data-analysis samples padding -vnext

dobromirts:


zdrawku (changes requested):
The Data Analysis with Chart data (data-analysis-dock-manager) sample is still broken:

image

zdrawku (changes requested):
Something is wrong with the dialog container, and this is not because of your changes, the problem is also reproducible on production:
1. Open grid-dynamic-chart-data/data-analysis
2. Select range of items and open the Chart view
3. Result:

image

It should be:
image

There should be a scrollbar of the list with chart images and the actual chart should be fully visible.

Could you please test the entire application user interaction design.

zdrawku (approved):

PR #2551, opened 2021-03-29
Fix data-analysis samples padding -master

dobromirts:


zdrawku (changes requested):
The selected item style is not applied. All items should be with not-selected style and upon selection to get a different color (without transparency). This issue is not present in the [example with DockManager](https://staging.infragistics.com/products/ignite-ui-angular/angular/components/general/data-analysis#data-analysis-with-dockmanager) you can check the styles there.
dim-issue

Note: Please perform the same fix for staging as well, as I have missed it

zdrawku (changes requested):
Now the chart preview window is broken:

image

zdrawku:
Also, could you please expand the panel with charts on the right (this relates to the data-analysis-dock-manager sample):

image

To be more like:

image

zdrawku (approved):

PR #2561, opened 2021-03-31
Replace node-sass with dart-sass -vnext

### LINKED ISSUE #2558: Start using 'dart-sass' instead of 'node-sass'
node-sass is deprecated, we should start using dart-sass instead.

image

dobromirts:
Closes #2558


zdrawku:
46 minutes build time .. previously it was 8 :(
image

src/app/data-entries/select/select-header-footer/select-header-footer.component.scss
```diff
@@ -73,11 +73,11 @@ $custom-button-group: igx-button-group-theme(
...
-    $item-background: (igx-color: ('grays', 300)),
-    $item-text-color: (igx-color: ('grays', 700)),
-    $item-hover-background: (igx-color: ('grays', 400)),
-    $item-hover-text-color: (igx-color: ('grays', 700)),
-    $item-border-color: (igx-color: ('grays', 700)),
+    $item-background: igx-color('grays', 300),
+    $item-text-color: igx-color('grays', 700),
+    $item-hover-background: igx-color('grays', 400),
+    $item-hover-text-color: igx-color('grays', 700),
+    $item-border-color: igx-color('grays', 700),
```
simeonoff, line 80: This is wrong, it should be:

```suggestion
    $item-background: igx-color($color: 'grays', $variant: 300),
    $item-text-color: igx-color($color: 'grays', $variant: 700),
    $item-hover-background: igx-color($color: 'grays', $variant: 400),
    $item-hover-text-color: igx-color($color: 'grays', $variant: 700),
    $item-border-color: igx-color($color: 'grays', $variant: 700),
```

simeonoff (changes requested):

simeonoff:
> 46 minutes build time .. previously it was 8 :(
> image

Something must be very wrong. You can expect some slowdown if the build process uses some wrapper to call the dart sass binary, however, not in the scale of 5x. We must investigate further before we merge this PR.

dobromirts:
After some research i found that, the the sass package on NPM is compiled to JS from Dart, so it has the same behavior as native Dart Sass (unlike the node-sass package, which is based on the deprecated libsass), just with a performance penalty due to running in JS (which shouldn't be a practical issue except for very large projects).
The pure JS version is slower than the stand-alone executable (https://sass-lang.com/dart-sass).
For projects that have a large codebase of sass files that need to be compiled and performance is important,
the DartVM executable version should be preferred or integrated in the project [cli-issue](https://github.com/angular/angular-cli/issues/15899). 
The option to use the fast DartVM executable version of dart-sass, should be possible with this library
https://github.com/sass/embedded-host-node as mention also in [here](https://github.com/webpack-contrib/sass-loader/issues/774).

dobromirts:
The issue with the build time increase is also mentioned in https://github.com/angular/components/issues/22676 , https://github.com/angular/angular-cli/issues/20792.

simeonoff (approved):

PR #2612, opened 2021-04-20
Update batch editing samples

### LINKED ISSUE #1775: Use scrollbarlove for all grid samples
Grids need to use the scrollbarlove instead of the default scrollbar, in order to it affected by any theme applied from the theming widget.

Related to https://github.com/IgniteUI/igniteui-docfx/issues/1783

### LINKED ISSUE #2476: Batch editing sample is confusing
### Description  
There's an additional empty component in the grid with transactions sample and I'm confused, whether this is the way I'm supposed to define grids with transactions.

### Steps to reproduce  

1. Go to https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/batch-editing
2. Check the html template
3. There's an `<app-grid-with-transactions>` component. This component is missing from the code example and without it transactions aren't turned on.

### Result  
Users cannot easily configure grids with transactions.

### Expected result  
Users should be able to get a complete example from the first sample they reach.

### Attachments  
N/A

wnvko:
Simplify batch editing samples and update to match docs.

Closes #[1775](https://github.com/IgniteUI/igniteui-docfx/issues/1775)
Closes #2476


hanastasov (approved):

src/app/tree-grid/tree-grid-batch-editing/tree-grid-batch-editing-sample.component.html
```diff
@@ -1,36 +1,39 @@
...
-            <igx-column field="ID" header="ID" dataType="number" width="200"></igx-column>
-            <igx-column field="Age" header="Age" dataType="number" width="120"></igx-column>
-            <igx-column field="Name" header="Full Name" dataType="string" width="240"></igx-column>
-            <igx-column field="Title" header="Title" dataType="string" width="180"></igx-column>
-            <igx-column field="HireDate" header="Hire Date" dataType="date" width="150"></igx-column>
-            <igx-column field="OnPTO" header="On PTO" dataType="boolean" width="80"></igx-column>
-            <igx-column filterable="false" width="150" [editable]="false" >
+        <igx-tree-grid igxPreventDocumentScroll #treeGrid [data]="data" primaryKey="ID" foreignKey="ParentID"
+            [width]="'100%'" [height]="'500px'" [rowEditable]="true">
+            <igx-column filterable="false" width="150" [editable]="false">
```
valadzhov, line 5: If there's width set the grid looks like this:
treegrid column
The DELETE button of the more nested children is not visible.
If the width is removed the button becomes visible even for deeply nested children:
tree grid nested children

wnvko: @simeonoff what you think about this sample. Should we set column width or could we move the delete and add button to the right end side of the grid. Note I am trying to simplify this sample as much as I can and focus on transactions - no templated cell, filtering, sorting or anything not related to the topic.

simeonoff: Can you use the actionstrip component to host the row actions?

wnvko: Yes, we can switch to actionstrip, but this will move off the focus of the sample. This is why IMO we should not use actionstrip.

Lipata (approved):

PR #2614, opened 2021-04-20
refactor(*): update themes -vnext

### LINKED ISSUE #2588: Review and update styles of LOB samples
The main samples had e refactoring that should be addressed on the LOB samples as well:
https://github.com/IgniteUI/igniteui-angular-samples/pull/2589

dobromirts:
Closes #2588


zdrawku (commented):
- There is an error with the Boston marathon sample which causes a problem with the loading:
image. The error is present on staging, you can use this PR to fix it.

- grid-finjs - on a second try to open the chart, the button is "disabled" and does not allow me show the chart. This problem might not be related to your code changes rather the refactoring that we did last month.
- columns collection - there is a log of space between the list items. You can use the grid-finjs-partial sample as a reference, same applies for :
image

zdrawku (approved):

PR #2629, opened 2021-04-23
fix(services): move data files from feature module to `app/data`

ViktorSlavov:
Move all data files from `services` feature module to `app/data` folder.

In the future, we would add any new existing data file to the data root, as this would lead to more reusability and minimum redundancy


live-editing/configs/ExportExcelConfigGenerator.ts
```diff
@@ -24,33 +24,33 @@ export class ExportExcelConfigGenerator implements IConfigGenerator {
 
         configs.push(new Config({
             component: 'ExcelExportSample1Component',
-            additionalFiles: ["/src/app/services/export-excel/data/invoiceData.ts"],
+            additionalFiles: ['/src/app/data/invoiceData.ts'],
```
gedinakova, line 27: Grid Excel Export sample is broken in stackblitz/codesandbox - not finding the data.

src/app/services/localization-samples/localization-all-resources/localization-all-resources.component.ts
```diff
@@ -13,7 +13,7 @@ import localeJA from '@angular/common/locales/ja';
 import localeKO from '@angular/common/locales/ko';
 import localeHans from '@angular/common/locales/zh-Hans';
 import localeHant from '@angular/common/locales/zh-Hant';
-import { DATA } from '../../data/nwindData';
+import { DATA } from '../../../data/nwindData';
```
gedinakova, line 16: This sample is also failing in stackblitz/codesandbox for missing data.

gedinakova (changes requested):

live-editing/configs/ExportExcelConfigGenerator.ts
ViktorSlavov: Thanks for catching that! I forgot that `shortenComponentPath` interferes w/ the import path on live editing. They should now be working properly

gedinakova (approved):

ViktorSlavov:
Abandoning this PR as all of these will be fixed w/ https://github.com/IgniteUI/igniteui-angular-samples/pull/2700

PR #2632, opened 2021-04-23
Update live editing and tree configs

HristoP96:


gulpfile.js
```diff
@@ -93,6 +93,7 @@ const processApp = (projectPath, dest, directoriesToExclude) => {
                     fs.readFile(file.path, 'utf-8', (err, content) => {
                         // Adjust sample application bundle files
                         const jsonObj = JSON.parse(content);
+                        let additionals = [];
```
ViktorSlavov, line 96: Should be `const`, since it's never reasigned

ViktorSlavov (approved):

PR #2650, opened 2021-04-29
feat(themes): theming example with angular material

SisIvanova:
Related to: IgniteUI/igniteui-docfx/issues/2726


src/styles.scss
```diff
@@ -1,3 +1,18 @@
...
+
+$my-app-theme: mat-light-theme((
+  color: (
+    primary: $my-app-primary,
+    accent: $my-app-accent,
+  )
+));
+
+@include angular-material-theme($my-app-theme);
+
```
simeonoff, line 15: Why do we need this in the global styles?

SisIvanova: If we remove the first two lines, the layout of the Material dropdown menus will break. I will delete the rest of the styles.

simeonoff: Why? Aren't you already including the core angular material styles as part of the sample?

src/app/theming/angular/angular-sample.component.ts
```diff
@@ -0,0 +1,65 @@
...
+import { FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { OverlayContainer } from '@angular/cdk/overlay';
+
+@Component({
+    selector: 'app-angular-sample',
+    styleUrls: ['./angular-sample.component.scss'],
+    templateUrl: './angular-sample.component.html'
+})
+
+export class AngularComponent implements OnInit {
```
simeonoff, line 12:
```suggestion
export class AngularMaterialComponent implements OnInit {
```

simeonoff (changes requested):

simeonoff (approved):

PR #2694, opened 2021-05-28
Optimize angular samples for cleaner and faster build 

### LINKED ISSUE #1410: Optimize angular samples for cleaner and faster build
### Description  
The repo's build time needs to be optimized in a way so that it could be faster.  The following things needs to be considered :

-  Json data repetition omission

-  CRM grid optimization

- Redundant css omission

- FinJS grids review

- Theme chooser sample review

zdrawku:
Closes #1410

Enable strict mode?


dobromirts (changes requested):
The removal of 'node --max_old_space_size' from these scripts is causing an error locally "'node_modules' is not recognized as an internal or external command" and they are not executed.

dobromirts (dismissed):

PR #2700, opened 2021-06-02
Restructure data services vNext

zdrawku:
Related to #1410


hanastasov (approved):
Good job!

Possible slight enhacements:
1. Employees.ts is same as employeesData.ts.
2. grid-editing-sample/data.ts is same as nWindData.ts
3. Possibly rename `src/app/grid/services/remote.service.ts` to `src/app/grid/services/remoteNwind.service.ts`. But there is another `remotenWind.service.ts` file, so this conflict should be resolved somehow.
4. and `export class RemoteService` to `export class RemoteNwindService` to be more descriptive
5. Replace all references to `RemoteService` with `RemoteNwindService` (Visual Studio Code should prompt to update it for us)

hanastasov (changes requested):
StackBlitz samples to be fixed:
- grid/remote-data-operations.html#remote-paging-with-batch-editing
- combo-remote.html

hanastasov (approved):

PR #2702, opened 2021-06-03
Remove grid with transactions wrapper component

### LINKED ISSUE #2476: Batch editing sample is confusing

wnvko:
Remove misleading grid with transactions components from the samples.

Closes #2476


Lipata (approved):

ViktorSlavov:
Since we're chaning the dialog content from a grid to a list, could you specify a max-height for the list item?
In the hierarchical grid demo, if you delete a parent, the list item representing the transaction can get a little out of hand:
image

kdinev:
Does the dialog grid necessarily have to go away?

wnvko:
@kdinev  as long as we are showing same `id`s in the dialog grid, and if `gridWithTransationsWrapper` is not there, transactions are applied to the dialog grid.
image
One way to fix this was to change it with list. We can also rethink the entire sample.

kdinev:
Doesn't the transaction service distinguish between grids the transactions are recorded for?

wnvko:
No. Transaction service has no knowledge about components. It is recording transaction against id provided from the user. Speaking of the grid it provides the id based on the row id. This is why if you have several grids with same row ids in a component you should wrap each grid in order to have each grid using its own transaction service.

kdinev:
This doesn't sound to me like a very good approach, since there could be multiple components, which may or may not collide in terms of record IDs. Having to explicitly force each component relying on transactions to be wrapped in another dummy component for the sake of injecting the service doesn't sound like the best possible idea.

wnvko:
Transaction service was implemented as no singleton service. The idea was each component to use its own transaction service for component's data. We can change this and store transactions for several components in single instance of transaction service, but at this moment this is a new product idea, and should be logged in separate issue.

src/app/grid/grid-batch-editing/grid-batch-editing-sample.component.scss
```diff
@@ -20,6 +20,7 @@ h4 {
     padding: 10px 0;
 }
 
+
```
wnvko, line 23:
```suggestion
```

src/app/grid/grid-paste/grid-paste-sample.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="grid__wrapper">
     <div class="grid__menu">
-        <button igxButton="raised" [igxToggleAction]="dropdown" [igxDropDownItemNavigation]="dropdown">Paste Options</button>
+        <button igxButton="raised" [batchEditing]="true" [igxToggleAction]="dropdown" [igxDropDownItemNavigation]="dropdown">Paste Options</button>
```
wnvko, line 3: `button` with `bathcEditing` - priceless 🤣 
please move this to the grid bellow

Lipata (approved):

PR #2725, opened 2021-06-16
Use toolbar for exporting samples, instead of button

### LINKED ISSUE #2723: Update Export to CSV and Excel samples to use the toolbar export feature
### Description  
Instead of having a button under the grid doing the export, enable the grids' export feature in the toolbar.
Apply this approach in all grids' samples for both Excel & CSV.

onlyexeption:
Closes #2723


gedinakova (changes requested):
By default, the two export options are set to true. We need to set the option, we don't want present, to false.

gedinakova (approved):

PR #2733, opened 2021-07-01
Remove localization samples configs shorten path -vnext

### LINKED ISSUE #2731: Localization topic samples are not working on SB and sandbox
https://www.infragistics.com/products/ignite-ui-angular/angular/components/general/localization

dobromirts:
Closes #2731


live-editing/configs/LocalizationConfigGenerator.ts
```diff
@@ -5,62 +5,57 @@ import {
...
-            shortenComponentPathBy: "/localization-samples/"
+            })
         }));
 
         // localization sample 3
         configs.push(new Config({
             component: 'LocalizationSample3Component',
-            additionalFiles: ["src/app/data/nwindData.ts"],
-            additionalDependencies: ["igniteui-angular-i18n"],
+            additionalDependencies: ['igniteui-angular-i18n'],
```
zdrawku, line 41: additionalFiles should not be removed as the sample in stackblitz will return an error. Is there a reason to remove adding the nwindData.ts?

zdrawku (changes requested):
I've noticed two problems, the first sample of the topic is not loaded because of missing package - its strange that the particular package is present ..
image

dobromirts:
> The first sample of the topic is not loaded because of missing package - its strange that the particular package is present ..
> image

This behaviour is due to the usage of  resources as 'IgxResourceStrings'  which are provided from this package and when they are assigned to 'locales' property in the 'OnInit' stackblitz can not find and use this package anymore.

zdrawku (approved):

PR #2756, opened 2021-07-28
Replace node-sass with dart-sass -master

### LINKED ISSUE #2558: Start using 'dart-sass' instead of 'node-sass'

dobromirts:
Closes #2558


zdrawku:
Already merged to staging

PR #2768, opened 2021-07-29
Add Range Buttons sample

### LINKED ISSUE #2750: Create an IgxDateRangePicker sample with pre-defined range button options
### Description  
As mentioned [here](https://www.infragistics.com/community/forums/f/ignite-ui-for-angular/123289/date-range-picker-with-pre-defined-range-button-options), we may want to create a sample which covers the scenario since it is a common one.

There can be multiple buttons on the sample where a user can select from pre-defined ranges. It could also be either in `dropdown` or `dialog` mode while making use of the `igxPickerActions` directive.

ddaribo:
Closes #2750


src/app/scheduling/daterangepicker/daterangepicker-range-buttons/daterangepicker-range-buttons.ts
```diff
@@ -0,0 +1,29 @@
...
+export class DateRangePickerRangeButtonsComponent {
+    @ViewChild('rangePicker', { static: true })
+    public dateRangePicker: IgxDateRangePickerComponent;
+
+    public range: DateRange = {
+        start: new Date(),
+        end: new Date(new Date().setDate(new Date().getDate() + 2))
+    };
+
+    constructor(public element: ElementRef) { }
```
jackofdiamond5, line 18: There's no need for the `element` to be public.

src/app/scheduling/daterangepicker/daterangepicker-range-buttons/daterangepicker-range-buttons.ts
```diff
@@ -0,0 +1,29 @@
...
+    public selectDays(count: number) {
+        let today: Date = new Date();
+        this.range = {
+            start: new Date(new Date().setDate(today.getDate() - count + 1)),
+            end: today
+        };
+
+        this.dateRangePicker.close();
+    }
+}
```
jackofdiamond5, line 29:
```suggestion
}

```

jackofdiamond5 (approved):

PR #2775, opened 2021-07-29
Fix colors for samples content and background in dark themes

### LINKED ISSUE #2763: Text content is not visible in many samples when a dark theme is applied.
### Description  
There seems to be a problem with the background in a lot of samples which causes the text content to be invisible when a dark theme is applied

### Steps to reproduce  

1. Navigate to [Text Highlight topic](https://www.infragistics.com/products/ignite-ui-angular/angular/components/texthighlight)
2. Apply any dark theme, e.g. "Material dark"

Note: Similar behavior is observed in most samples under "Data entry & Display" & "Layouts" categories

### Result  
The text disappears

### Expected result  
Text content should be visible.

### Attachments  
image
image

desig9stein:
Closes #2763

### DIFF
--- a/projects/app-lob/src/app/index/docs-layout.component.scss
+++ b/projects/app-lob/src/app/index/docs-layout.component.scss
@@ -1,6 +1,6 @@
 ::ng-deep {
-    .custom-body {
-        color: var(--igx-grays-800);
-        background-color: var(--igx-background-500);
-    }
+	.custom-body {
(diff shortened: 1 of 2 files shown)


simeonoff (approved):

gedinakova:
@simeonoff @desig9stein How can theming be tested locally? The widget doesn't seem to work when I run it?

PR #2778, opened 2021-07-29
bind to cell in template

hanastasov:


dobromirts:
On scroll "Cannot read property 'data' of undefined" error is thrown in the console.

dobromirts (approved):

PR #2837, opened 2021-11-04
chore(*): adding gh actions yml

ChronosSF:


kdinev:
@ChronosSF The GH actions builds are now three. I was kind of asking for just one :)

ChronosSF:
> @ChronosSF The GH actions builds are now three. I was kind of asking for just one :)

With the latest changes I left 2 - for node 14 and node 16. I could only leave 16 if you prefer. I guess I could also delete azurepipelines.yml and disable the run from the branch protection rules (afaik that's where it triggers from)

kdinev:
@zdrawku Were the two old builds used for anything?

ChronosSF:
> @zdrawku Were the two old builds used for anything?

It was setup like this:

- azure pipes running always but building just the non-lob , non-crm build
- the crm build running only for changes in the crm dir
- the lob build running only for changes in the lob dir

It is a better setup most of the time, as long as the three apps are fully isolated from each other. With a single build I run all 3 builds for each change, regardless where it is.

kdinev:
> > @zdrawku Were the two old builds used for anything?
> 
> It was setup like this:
> 
> * azure pipes running always but building just the non-lob , non-crm app
> * the crm GH actions building just the crm app and running only for changes in the crm dir
> * the lob GH actions building just the lob app and running only for changes in the lob dir
> 
> It is a better setup most of the time, as long as the three apps are fully isolated from each other (i.e. they don't import from each other). With a single build I run all 3 builds for each change, regardless where it is.

If any of the builds fail, then the PR is no good, so separating the builds doesn't make a whole lot of sense to me. @zdrawku?

ChronosSF:
> > > @zdrawku Were the two old builds used for anything?
> > 
> > 
> > It was setup like this:
> > 
> > * azure pipes running always but building just the non-lob , non-crm app
> > * the crm GH actions building just the crm app and running only for changes in the crm dir
> > * the lob GH actions building just the lob app and running only for changes in the lob dir
> > 
> > It is a better setup most of the time, as long as the three apps are fully isolated from each other (i.e. they don't import from each other). With a single build I run all 3 builds for each change, regardless where it is.
> 
> If any of the builds fail, then the PR is no good, so separating the builds doesn't make a whole lot of sense to me. @zdrawku?

It does if the three apps are fully isolated and don't import from each other. When the builds are separated in this way, it allows for just the build of the modified code to be executed saving on build time. We use this approach in other places as well.

However, I am not sure if the apps are isolated so I'll leave @zdrawku  to give us a clue.

PR #2843, opened 2021-11-08
feat(action-strip): add description to tree and hgrid component

Svetloslav15:
This PR is related to this issue: [igniteui-docfx/3007](https://github.com/IgniteUI/igniteui-docfx/issues/3007)


hanastasov (changes requested):
Good job! 

After opening the sample in stackblitz, the grid is empty. Maybe it cant find the data. If data is in addittionall file, this filer should be included in the live-editing config. See readme here https://github.com/IgniteUI/igniteui-live-editing

PR #2863, opened 2021-11-17
Restore sass imports & update igniteui-angular & @igniteui/angular-schematics

dobromirts:


Lipata:
@hanastasov, @jackofdiamond5 see the [commit ](https://github.com/IgniteUI/igniteui-angular-samples/pull/2863/commits/7f0e76c55ead62c526e6d5b811fd4182f7f26bec) above. I don't think that we can easily add migration for such cases, so they need to be documented?

Lipata (approved):

PR #2875, opened 2021-11-19
style(*): Use hsla for surface to apply proper background when theme is changed

dkamburov:


projects/app-lob/src/app/index/docs-layout.component.scss
```diff
@@ -1,6 +1,6 @@
 ::ng-deep {
 	.custom-body {
-		color: var(--igx-surface-500-contrast);
-		background: var(--igx-surface-500);
+		color: hsla(var(--igx-surface-500-contrast));
```
wnvko, line 3: `hsla` function should not be applied to contrast colors

src/app/index/docs-layout.component.scss
```diff
@@ -1,6 +1,6 @@
 ::ng-deep {
 	.custom-body {
-		color: var(--igx-surface-500-contrast);
-		background: var(--igx-surface-500);
+		color: hsla(var(--igx-surface-500-contrast));
```
wnvko, line 3: `hsla` function should not be applied to contrast colors

wnvko (changes requested):

zdrawku (changes requested):
@simeonoff @wnvko @dkamburov 

Why we dont apply hsla to contrast colors?

Also, can we update all css palette variables to use the hsla css function

image

simeonoff:
> @simeonoff @wnvko @dkamburov
> 
> Why we dont apply hsla to contrast colors?
> 
> Also, can we update all css palette variables to use the hsla css function
> 
> image

Contrast colors are always either black or white - no need. No, the way it is right now allows you to pass the values to either the HSL or the HSLA function.

simeonoff (dismissed):

wnvko (dismissed):

zdrawku:
@dkamburov does the problem with the backgroud apper to ther sample as well, not only the docs-layout, but toggle, calendar and multiview?

image

dkamburov:
> @dkamburov does the problem with the backgroud apper to ther sample as well, not only the docs-layout, but toggle, calendar and multiview?
> 
> image

This change is related to the background of all samples but if you are ok to I can change the surface for those specific sample styles as well.

zdrawku:
> > @dkamburov does the problem with the backgroud apper to ther sample as well, not only the docs-layout, but toggle, calendar and multiview?
> > image
> 
> This change is related to the background of all samples but if you are ok to I can change the surface for those specific sample styles as well.

If the same problem persists there as well, please do the change, otherwise just dont :)

PR #2876, opened 2021-11-22
Update simple combo samples

wnvko:
Remove unnecessary console logs.
Update method names.
Add outlet to styling sample.
Update SCSS to use `use` instead of `import`.


Lipata (approved):

src/app/lists/combo/simple-combo-main/simple-combo-main.component.ts
```diff
@@ -7,7 +7,7 @@ import { StockData, updatedStockData } from '../../../data/stocks-data';
     templateUrl: 'simple-combo-main.component.html',
     styleUrls: ['simple-combo-main.component.scss']
 })
-export class SimpleComboMainComponent {
+export class SimpleComboMainComponent implements OnInit {
```
jackofdiamond5, line 10: We may want to apply some styling to the simple combo's element. I'm thinking `flex-grow: 1` and `max-width: 300px` for example. This should prevent the combo from changing its size when there's text in the input field due to the addition of the clear icon.

Lipata (approved):

PR #2903, opened 2021-12-06
fix(grid): cond row selection fix for deselect all

kdinev:
Related https://github.com/IgniteUI/igniteui-angular/issues/10689


IvayloG:
Sample works prior to merging to vNext

IvayloG (changes requested):
The same fix should be done and will work for the hgrid and tgrid samples as well:
- hierarchical-grid-conditional-row-selectors.component.ts
- tree-grid-conditional-row-selectors.component.ts

kdinev:
> 

@IvayloG Done!

IvayloG (approved):

PR #2920, opened 2022-02-09
chore(toast): update toast samples to use positionSettings

SisIvanova:
Related to IgniteUI/igniteui-angular/issues/10947


MayaKirova:
Shouldn't this PR target vNext, since that's the version which will fist be updated to the 13.1.0? We could also update master, but vNext is where we actually need this in order to successfully update the igniteui version to 13.1.0 and have working samples.

simeonoff:
> Shouldn't this PR target vNext, since that's the version which will fist be updated to the 13.1.0? We could also update master, but vNext is where we actually need this in order to successfully update the igniteui version to 13.1.0 and have working samples.

@MayaKirova I changed the base to vNext.

simeonoff (approved):

PR #2922, opened 2022-02-10
feat(pivot): Add pivot noop sample

mddragnev:


dkamburov:
@mddragnev live editing generation fails, seems like a wrong dependency path

MayaKirova (approved):

dkamburov:
image

The first dimension shouldn't take 25%, maybe add more data for one or two more columns.

ChronosSF:
@mddragnev , please use super compact mode. Also if we have 0 for the empty fields i.e. 0 units sold it will probably look better

ChronosSF (changes requested):
@mddragnev , please use super compact mode. Also if we have 0 for the empty fields i.e. 0 units sold it will probably look better

PR #2928, opened 2022-02-17
Data Selector Sample

### LINKED ISSUE #2924: Pivot Grid sample basic with pivot data selector

3phase:
Closes #2924


src/app/pivot-grid/pivot-datasource/pivot-datasource-sample.component.ts
```diff
@@ -0,0 +1,60 @@
...
+    templateUrl: './pivot-datasource-sample.component.html'
+})
+export class PivotDatasourceSampleComponent {
+    public data = DATA_TO_ANALYZE;
+    public pivotConfigHierarchy: IPivotConfiguration = {
+        columns: [
+
+            {
+                memberName: 'Region',
+                memberFunction: (data) => data.Region,
```
dkamburov, line 18: I don't think that the `memberFunction` is needed for this scenario

MayaKirova:
The pivot items have a lower zindex than the grid so when you drag them near the grid they disappear behind it:
list-items-zindex

I can't reproduce the same in the dev samples, so there's probably some styling in the demos that causes this.

MayaKirova (dismissed):

dkamburov:
@3phase The height of the grid should match the data selector, also there is no live-editing configuration for the sample

MayaKirova:
@3phase @dkamburov 

- [ ] The sample does not fit in the size available in the docs. Either it should be smaller or the size in the docs should be updated to match:

image

- [ ] Pivot selector z-index here seems to be lower than the grid. 

image

- [ ] We should disable column selection similar to the features sample, since it's not a pivot specific feature.

- [ ] The current way the dimensions are organized will always create empty values since specific Dates are grouped in the rows while the Years are grouped in the columns. So specific dates will always have a whole column section with no values (i.e. all 2019 dates will have empty value when you scroll to the 2018 group and vice versa).
image

It would make more sense to group date related values in one dimension so that you see meaningful aggregations for it.
For example rows can contain some date hierarchy (year -> month -> date), and everything else can be grouped in the columns or vice versa.

MayaKirova (changes requested):
@3phase @dkamburov 

- [x] The sample does not fit in the size available in the docs. Either it should be smaller or the size in the docs should be updated to match:

image

- [ ] Pivot selector z-index here seems to be lower than the grid. 

image

- [x] We should disable column selection similar to the features sample, since it's not a pivot specific feature.

- [x] The current way the dimensions are organized will always create empty values since specific Dates are grouped in the rows while the Years are grouped in the columns. So specific dates will always have a whole column section with no values (i.e. all 2019 dates will have empty value when you scroll to the 2018 group and vice versa).
image

It would make more sense to group date related values in one dimension so that you see meaningful aggregations for it.
For example rows can contain some date hierarchy (year -> month -> date), and everything else can be grouped in the columns or vice versa.

dkamburov:
The size of sample in topic is updated in - https://github.com/IgniteUI/igniteui-docfx/pull/3130

3phase:
@MayaKirova the requested changes are applied

dkamburov:
@3phase 
The filters from data selector are cut off
image

MayaKirova (approved):

PR #2972, opened 2022-03-15
Updates hierarchical grid config to include reference to models

3phase:


mtsvyatkova (dismissed):

MayaKirova:
The last stackblitz config seems to be missing the correct data file:
HGridActionStripSampleComponent

It seems to also have reference to the singers data source, but does not refer it in the stackblitz config:
image

It probably should have reference to singersData and the model.

3phase:
@MayaKirova it's done now, thanks for noticing

skrustev (approved):

PR #2987, opened 2022-03-31
chore(*): Fix horizontal sample orientation.

### LINKED ISSUE #2985: Horizontal virtualization sample has incorrect layout.
### Description  
Layout should be horizontal as it shows horizontal virtualization, but elements are ordered vertically.

### Steps to reproduce  

1. Go to the sample at: https://www.infragistics.com/products/ignite-ui-angular/angular/components/for-of#horizontal-virtualization
2. Observe

### Result  
List items are ordered vertically and horizontal scrollbar is not even visible.

### Expected result  
Elements should display in horizontal layout and the horizontal scrollbar should be visible.

MayaKirova:
Closes #2985


3phase:
The z-index of the focused item is getting bigger than their sibling, so the following happens:

image

I think that we should either change the z-index to 2, or investigate what causes the wider spanning item.

Besides that all looks well 😊

MayaKirova:
@dkamburov Content seems to overflow the 220px set for the item when active. Size should be fixed to 220px for all items now.

dkamburov (approved):

PR #3030, opened 2022-06-10
chore: remove --force

Lipata:


Lipata:
Still waiting for official @ng-bootstrap/ng-bootstrap that is angular 14 compatible. Now using beta, which is still not stable: https://github.com/ng-bootstrap/ng-bootstrap/issues/4340

zdrawku (approved):

PR #3033, opened 2022-06-13
adding popperjs to deps for prod build - vnext

ChronosSF:


damyanpetev (approved):
@popperjs/core is a peer dep of newer version of ng-bootstrap so this seems correct

PR #3041, opened 2022-06-23
docs(grid): add sample for excel style editing

tishko0:
Sample for #3238


src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      const columnName = grid.getColumnByVisibleIndex(activeElem.column).field;
+      const cell = grid.getCellByColumn(activeElem.row, columnName);
+      if (cell && !grid.crudService.cellInEditMode) {
+        grid.crudService.enterEditMode(cell);
+        cell.editValue = event.key;
+      }
+    }
+
+    if (key == 13) {
```
MayaKirova, line 49: @dkamburov Should we also do Shift+Enter like in excel for this sample? Might as well have both directions.

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
```diff
@@ -0,0 +1,87 @@
...
+      const maxRows = rowInfo.length;
+
+      if (rowInfo[thisRow + 1]) {
+        while ('expression' in rowInfo[thisRow + 1]) {
+          ++thisRow;
+          ++nextRow;
+        }
+      }
+      if (nextRow < maxRows) {
+        requestAnimationFrame(() => {
```
MayaKirova, line 65: Is this requestAnimationFrame needed for something? I don't think we need to do this asynchronously.

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
```diff
@@ -0,0 +1,87 @@
...
+            this.grid.navigateTo(nextRow, column);
+            this.grid.tbody.nativeElement.focus();
```
MayaKirova, line 67: Is this focus needed?

MayaKirova (changes requested):

MayaKirova:
Also a config should be added to live-editing\configs\GridConfigGenerator.ts for this sample so that it can generate a stackbltz sample.

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
```diff
@@ -0,0 +1,87 @@
...
+@Component({
+  selector: 'app-grid-excel-style-editing-sample',
+  styleUrls: [`grid-editing-excel-style.component.scss`],
+  templateUrl: 'grid-editing-excel-style.component.html'
+})
+export class GridExcelStyleEditingComponent implements OnInit {
+  @ViewChild('grid', { read: IgxGridComponent, static: true })
+  public grid: IgxGridComponent;
+  @ViewChild(IgxDialogComponent, { static: true })
+  public dialog: IgxDialogComponent;
```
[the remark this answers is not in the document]
tishko0: removed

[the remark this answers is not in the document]
tishko0: removed

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
```diff
@@ -0,0 +1,87 @@
...
+      this.grid.endEdit();
+      this.grid.clearCellSelection();
+      let thisRow = activeElem.row;
+      let nextRow = thisRow + 1;
+      const column = activeElem.column;
+      const rowInfo = grid.dataView;
+      const maxRows = rowInfo.length;
+
+      if (rowInfo[thisRow + 1]) {
+        while ('expression' in rowInfo[thisRow + 1]) {
```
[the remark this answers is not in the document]
tishko0: not sure we have such method that is not private, I can add check for summary/master detail instead?

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
tishko0: keeps the focus on the grid, doesn't really work without it for some reason. Stackblitz sample does tho

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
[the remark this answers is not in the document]
MayaKirova: IMO add  a `getNextEditableRowIndex` method in the sample that follows similar logic to the internal [getNextDataRowIndex](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts#L7246) method of the grid. It should contain check for all non-data rows, like in  [isEditableDataRecordAtIndex](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts#L7284)

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
MayaKirova: Looks like some sort of a workaround. If something needs to happen asynchronously it probably needs to happen after navigation is done. You can try moving the activation and selection of the node in the `navigateTo` callback method:
```
this.grid.navigateTo(nextRow, column, () => {
// anything that needs to happen after the cell comes in view goes here
});
```

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
dkamburov: Yes, that'd be great.

src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (key == 13) {
+        this.grid.clearCellSelection();
+        let thisRow = activeElem.row;
+        const column = activeElem.column;
+        const rowInfo = grid.dataView;
+
+        let nextRow = this.getNextEditableRowIndex(thisRow, rowInfo, event.shiftKey);
+
+        this.grid.navigateTo(nextRow, column, (obj) => {
+          obj.target.activate();
```
MayaKirova, line 53: It seems that here the active element after enter is lost, seems to be a change detection cycle missing, so if you do:
```
 this.grid.navigateTo(nextRow, column, (obj) => {
          obj.target.activate();
          this.grid.cdr.detectChanges();
          });
      }
```

It should be fine.

MayaKirova (changes requested):

MayaKirova:
@tishko0 
You could also disable multi-cell selection so that you don't have to clean selections on Shift+Enter:
[cellSelection]="'single'"

MayaKirova (approved):

PR #3045, opened 2022-06-27
Ttonev/grid excel style editing vnext

tishko0:
Sample for #3238


src/app/grid/grid-editing-excel-style/grid-editing-excel-style.component.html
```diff
@@ -0,0 +1,48 @@
+<div class="grid__wrapper">
+    <igx-grid
+      igxPreventDocumentScroll
+      #grid
+      [data]="data"
+      [primaryKey]="'ProductID'"
+      width="100%"
+      height="500px"
```
mddragnev, line 8: Please, run both the docfx and the samples and observe that there is much white space between the live editing buttons and the sample itself
Screenshot 2022-07-05 at 17 34 06

tishko0: I've adjusted the height values so the sample fits in the designated window, in case it should be bigger, let me know

dkamburov:
@tishko0 I'd be great to handle arrow keys while in editing

dkamburov (dismissed):

dkamburov (approved):

dkamburov (approved):

PR #3051, opened 2022-07-04
fix(grid-selection):Updating the use of rowSelectionChanging event in…

### LINKED ISSUE #3042: Grid rowSelectionChanging event samples issue
(rowSelectionChanging) event samples don't look right.
 - It's not typed
 - There doesn't seem to be such thing as event.cell
 - The sample doesn't do anything with it, so it's not showing how this event is useful.

image

IvanKitanov17:
Closes #3042


zdrawku (changes requested):
Well done with the idea. I have a few suggestions in order to improve this:
- Can you consider making the snackbar height dynamic, based on its content. This new snackbar should be with smaller height than the others as the content is a one-liner. Also please use the same concept with the "bell" icon and action button to close the snackbar "got it, thanks"
image
- Can you change this action to address the "Currently selected row index" instead of number of selected rows, as the numbers wont change when the selection is set to "single" and the former will make more sense than the latter
image

- When the changes above are addressed, please apply them to the hGrid and TreeGrid selection samples
image

zdrawku (approved):

PR #3069, opened 2022-08-17
Ttonev autosizing sample

tishko0:
Added samples for the new autoresize feature of the grids


tishko0:
The current version of vNext does not have the latest changes as this feature is introduced at 14.1, so the samples will not work yet. There seems to be a style missing from the latest build that is not yet into vNext, so I suppose some migrations will need to be done before the samples can be tested in this state.

dkamburov:
@tishko0 please resolve the conflicts

live-editing/configs/app-dv-configs/DVHierarchicalGridConfig.ts
```diff
@@ -28,6 +28,21 @@ export class DVHierarchicalGridConfigGenerator implements IConfigGenerator {
             component: 'HGridColumnResizingSampleComponent'
         }));
```
ChronosSF, line 30: you probably wanted to modify live-editing/configs/HierarchicalGridConfigGenerator.ts

dkamburov:
@tishko0 The build fails, there must be an issue with the merge

ChronosSF (changes requested):

dkamburov:
@tishko0 currently hierarchical grid sample is not loading and throwing an error.

live-editing/configs/app-dv-configs/DVHierarchicalGridConfig.ts
tishko0: yes, moved now

tishko0:
> 

fixed now

ChronosSF:
@tishko0 - some observations and requests

1. always format the files before committing
2. remove things copy pasted from other samples that are not needed
3. make the samples focus on the feature shown, remove the fluff coming from samples you derive yours from
4. always make sure that `npm run start:live-editing` passes
5. don't include LOB stuff if not needed (such as the sparkline in the hierarchical sample)
6. test your samples with a running docfx app so that you can see if your generation configs work (by clicking on the stackblitz button)
7. ideally make a self-review on the PR . You were deleting two routes for two other samples that made live-editing to fail

ChronosSF (approved):

PR #3077, opened 2022-08-23
refactor(samples): changes with igniteui-theming

simeonoff:


src/_variables.scss
```diff
@@ -9,8 +9,11 @@ $default-palette: $palette;
 $green-palette: palette(
     $primary: #09f,
     $secondary: #72da67,
-    $grays: #fff,
-    $surface: #333
+    $surface: #333,
```
didimmova, line 12: You're trying to get the gray color shade from this palette but there isn't any gray color passed to the palette.

simeonoff: That's fine, the gray color is automatically generated when not provided based on the surface color.

didimmova:
- In shadows/card-sample-shadow.scss file the elevations function is being used and it doesn’t exist anymore in the theming package. Are we going to create a function like this there so we can keep using it?

- No surface color in palettes in app/theming/styles files

simeonoff:
@didimmova All version references here need to be updated from 14.2.x. to 15.0.x.

didimmova:
Fix the custom elevations color sample in shadows/card-sample-shadow.scss file

simeonoff:
> 

Done.

simeonoff:
> Fix the custom elevations color sample in shadows/card-sample-shadow.scss file

Done.

didimmova (dismissed):

desig9stein (approved):

PR #3079, opened 2022-08-24
Add pivot state persistance sample

mddragnev:


MayaKirova:
@mddragnev 
The "Go Back" button does not work, it seems that it has a different url that does not match the sample:
Cannot match any routes. URL Segment: 'samples/pivot-grid/pivot-grid-state-persistance'

mddragnev:
@MayaKirova 
Both things should be okay now. Do you think that we need to remove row selection from the topic as well, or save it there and just fix the issue?

MayaKirova:
@mddragnev think it can stay and we can enable it back in the sample once the issue is fixed.

tishko0 (approved):

mddragnev:
> 

@dkamburov I do not encounter that error. On my machine everything works fine. I have removed node_modules and reinstalled them. Both of the samples load correctly in the docfx.

dkamburov:
I could've had some leftovers in the local storage

PR #3082, opened 2022-08-30
docs(validator): Add sample for styling the grid validator

dkamburov:


MayaKirova (approved):

simeonoff (dismissed):

mddragnev:
@dkamburov In the tree grid for some reason row selection with rowStyles does not work, we should better remove row selection.

mddragnev:
@dkamburov So I have investigated this further and what happens is that this is not an issue but expected behavior. The thing we can do if you want to have row selection + row styles, we should add `rowClasses` instead of `rowStyles`.

dkamburov:
@mddragnev Ok, great, but it seems a bit better to remove the row selection, no need to pollute the row and show off that feature here

MayaKirova (dismissed):

MayaKirova (approved):

PR #3083, opened 2022-08-31
Add cross cell validation sample.

skrustev:


MayaKirova (commented):
Looks good. Only thing I noticed that I'm not sure if intended is that it collects only the custom errors, the default ones (like required, email etc) do not show in the pinned cell.

image

src/app/grid/grid-validator-service-cross-cell/grid-validator-service-cross-cell.component.ts
```diff
@@ -0,0 +1,170 @@
...
+    @ViewChild('grid1', { read: IgxGridComponent })
+    public grid: IgxGridComponent;
+
+    public minDealsRatio = 0;
+    public transactionData = JSON.parse(JSON.stringify(employeesData));
+
+    constructor() {
+    }
+
+    public formCreateHandler(formGroup: FormGroup) {
```
MayaKirova, line 84: We changed the event arg to IGridFormGroupCreatedEventArgs. This will need update.

MayaKirova (changes requested):

MayaKirova (approved):

mddragnev:
@skrustev 
One thing I found is in the tree grid, you have set a max validator on the `age` column but there is not a handler in the stateMessage for that validator. Once this validator is invalid then the row valid column count is as invalid but there is no message.
And a side note, in the tree grid sample if we pin the row valid column on the right, IMO we should make the width of the grid (or width of some of the columns) should be less because the tooltip is getting cut at the end.

MayaKirova (approved):

PR #3085, opened 2022-09-07
Add cross-field validation sample for Hierarchical grid.

MayaKirova:


mddragnev:
@MayaKirova If you try to make any changes in any cell of the child grid and then press commit, the data of the child grid refresh to the initial state.
Also, on the main grid, if you set country to be in validation error with the city and commit the message, all the validation errors are cleared (and that is okay) but after you enter edit mode again for the exact same cell, unless you make a new change from the select you dont see error.

mddragnev:
> @mddragnev Regarding the cleared state. Once the validation is cleared the state of the data is considered 'pristine' so until anything is touched explicitly by the user no errors will be shown. Same as if there's initially invalid data in the grid.

Thats true but if you enter edit mode again on that same cell it does not refresh the errors state

MayaKirova:
> > @mddragnev Regarding the cleared state. Once the validation is cleared the state of the data is considered 'pristine' so until anything is touched explicitly by the user no errors will be shown. Same as if there's initially invalid data in the grid.
> 
> Thats true but if you enter edit mode again on that same cell it does not refresh the errors state

Hm, that is strange indeed. I'll take a look.

skrustev (approved):

dkamburov (approved):

PR #3091, opened 2022-09-12
Remove some code from the old API. Fix tree grid cross-field cells wi…

mddragnev:
…th select component. Some more issues


MayaKirova:
1. The phoneFormat validator at [hierarchical-grid-validator-service-extended.component.html](https://github.com/IgniteUI/igniteui-angular-samples/pull/3091/files#diff-85c782c08f677404637be2e4d89ffe940f64dd436686df20a300ccdbd4df789d) never seems to be invoked so the error never shows.

2. I see that a new import was added at: https://github.com/IgniteUI/igniteui-angular-samples/pull/3091/files#diff-508cbce15cb0b9ae2284501fcea09c595e5df582c7030534f1b5a907122b4cf6R236

   Same import should be added in the Stackblitz config as well. Otherwise it will not work.

mddragnev:
@MayaKirova
1. For the first question it does invoke but you need to put a symbol different than a digit. For example try adding 'abv' in the end.

live-editing/configs/HierarchicalGridConfigGenerator.ts
```diff
@@ -993,9 +993,9 @@ export class HierarchicalGridConfigGenerator implements IConfigGenerator {
 
             ],
             appModuleConfig: new AppModuleConfig({
-                imports: ['HierarchicalGridValidatorServiceExtendedComponent', 'IgxHierarchicalGridModule', 'IgxPreventDocumentScrollModule'],
+                imports: ['HierarchicalGridValidatorServiceExtendedComponent', 'IgxHierarchicalGridModule', 'IgxPreventDocumentScrollModule', 'ReactiveFormsModule'],
```
MayaKirova, line 996: Shouldn't this be added in the TreeGridConfigGenerator for the cross-field validation sample config?

MayaKirova:
@mddragnev 
1. Seems to have been an issue on my side. Works ok now.
2. Seems ok but shouldn't it be in the tree grid config? Since the modules for the tree grid were the ones updated.

live-editing/configs/HierarchicalGridConfigGenerator.ts
mddragnev: Yup, I have misplaced it. Gonna fix that now.

mddragnev:
> @mddragnev
> 
> 1. Seems to have been an issue on my side. Works ok now.
> 2. Seems ok but shouldn't it be in the tree grid config? Since the modules for the tree grid were the ones updated.

I tested it now. Should be working.

MayaKirova (approved):

PR #3127, opened 2022-10-18
adding dark parameter handling for crm grid

ChronosSF:


ChronosSF:
@desig9stein - to test this use http://localhost:4200/?dark=true . Also ensure just runnign http://localhost:4200 has the grid style untouched compared to before the change.

PR #3128, opened 2022-10-19
Add Simple Combo remote binding sample and filtering to Combo remote binding sample

### LINKED ISSUE #3076: IgxSimpleCombo remote binding sample
### Description  
We should add a remote binding sample for our `IgxSimpleComboComponent` and include it in the docs as well. Right now we only have a remote binding sample for the multi select combo. A new sample for the simple combo is needed because while the components are similar in API and behavior, for the most part, there are a few differences that that need to be addressed when binding to remote data, so we cannot simply substitute the `igx-combo` tag for the simple combo's tag.

Related to https://github.com/IgniteUI/igniteui-angular/issues/11885

Enable filtering in the combo remote sample, because now it is disabled.

IvanKitanov17:
Closes #3076


jackofdiamond5:
I found two little problems with the combo and simple combo - if you apply filtering and select a filtered element then when you reopen the combo it will still act as if it has applied filtering and won't load all of the data.

simple-combo-remote

combo-remote

jackofdiamond5 (dismissed):

RivaIvanova:
I found a strange behavior when performing the following steps:
1. Select an item and scroll down until it is not visible.
2. Close the combo and open it again - The scroll position is set so that the last selected item is visible which seems expected.
3. Try scrolling up - the data starts loading from the last chunk that was visible before closing the combo.

combo-remote

The same should be true for the simple combo as well since the implementation seems the same, however, I noticed an [issue](https://github.com/IgniteUI/igniteui-angular/issues/12289) when having a simple combo bound to remote data and I wasn't able to test it.

jackofdiamond5 (approved):

PR #3150, opened 2022-11-07
removing the dark theme switch

ChronosSF:
We are removing the switch because the finjs sample can now be styled with the theming widget.


hanastasov (approved):

hanastasov (commented):
Also, all occurences of `fin-dark-theme` can be removed. and also remove the theme from tree grid finjs sample

hanastasov (approved):

Issue #3155, opened 2022-11-16
Add loading indicator for cascading combo sample


Lipata:
### Description  
Go to the simple combo cascading sample: https://www.infragistics.com/products/ignite-ui-angular/angular/components/simple-combo#cascading-scenario

### Expected result  
As an improvement, a loading indicator can be added for 2nd and 3rd combo - from selecting an item in the previous combo, until next combo data is loaded.

PR #3159, opened 2022-11-21
Add loading indicator for cascading simple combo sample vNext

georgianastasov:
Closes: #3155

### DIFF
--- a/src/app/lists/combo/simple-combo-cascading/simple-combo-cascading.component.ts
+++ b/src/app/lists/combo/simple-combo-cascading/simple-combo-cascading.component.ts
@@ -25,6 +25,7 @@ export class SimpleComboCascadingComponent implements OnInit {
     public countryChanging(e: ISimpleComboSelectionChangingEventArgs) {
         this.selectedCountry = e.newSelection as Country;
         if (e.newSelection){
+            this.regionData = [];
             this.isLoadingRegions = true;
             this.loadingTime = 2000;
         }
@@ -43,6 +44,7 @@ export class SimpleComboCascadingComponent implements OnInit {
     public regionChanging(e: ISimpleComboSelectionChangingEventArgs) {
         this.selectedRegion = e.newSelection as Region;
         if (e.newSelection){
+            this.citiesData = [];
             this.isLoadingCities = true;
             this.loadingTime = 2000;
         }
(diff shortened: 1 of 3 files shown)


src/app/lists/combo/simple-combo-cascading/simple-combo-cascading.component.ts
```diff
@@ -14,24 +14,44 @@ export class SimpleComboCascadingComponent implements OnInit {
...
+    private loadingTime = 0;
+
     public ngOnInit(): void {
         this.countriesData = getCountries(['United States', 'Japan', 'United Kingdom']);
     }
 
     public countryChanging(e: ISimpleComboSelectionChangingEventArgs) {
         this.selectedCountry = e.newSelection as Country;
-        this.regionData = getCitiesByCountry([this.selectedCountry?.name])
+        if(e.newSelection){
```
Lipata, line 27:
```suggestion
        if (e.newSelection){
```

src/app/lists/combo/simple-combo-cascading/simple-combo-cascading.component.ts
```diff
@@ -14,24 +14,44 @@ export class SimpleComboCascadingComponent implements OnInit {
...
         this.selectedRegion = null;
         this.selectedCity = null;
         this.citiesData = [];
+        this.loadingTime = 0;
     }
 
     public regionChanging(e: ISimpleComboSelectionChangingEventArgs) {
         this.selectedRegion = e.newSelection as Region;
-        this.citiesData = getCitiesByCountry([this.selectedCountry?.name])
+        if(e.newSelection){
```
Lipata, line 45:
```suggestion
        if (e.newSelection){
```

Lipata (approved):

Lipata (dismissed):

RivaIvanova:
When initially selecting a country from the first combo, the region combo is disabled, because it does not have any data yet and the loading indicator is shown (as expected). However, if I select another country, the second combo is enabled as it is still bound to the previous data, which allows me to select a region that is not from the selected country while the loading indicator is still in progress. The same is true for the city combo when selecting another region.

Although the selected region is not a valid one, the city combo shows its loading indicator since a selection was made and at some point both indicators are visible.

cascading-combo

Perhaps setting `this.regionData` and `this.citiesData` to empty arrays before the `setTimeout` function for the respective handlers could solve this.

Lipata (approved):

Issue #3182, opened 2022-11-24
Update the Hierarchical Grid and Tree Grid Custom Column Pinning UI samples with the correct pinning icon


viktorkombov:
### Description  
Update the [Hierarchical Grid ](https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/column-pinning#demo-1) and [Tree Grid](https://www.infragistics.com/products/ignite-ui-angular/angular/components/treegrid/column-pinning#demo-1) Custom Column Pinning UI samples to use the same pinning icon as the [Data Grid Custom Column Pinning UI sample](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-pinning#demo-1)
<img width="51" alt="Screenshot 2020-04-02 at 15 43 21" src="https://user-images.githubusercontent.com/1702064/78250796-1539a700-74f9-11ea-9133-6746bc5c3fbe.png">

viktorkombov:
Related to #1898

PR #3184, opened 2022-11-24
fix(grids): improve the appearance of the pinning icon in the Custom Column Pinning UI samples

### LINKED ISSUE #3163: Update the Custom Column Pinning UI sample so that the pin icon to be always visible
### Description  
Update the Custom Column Pinning UI sample so that the pin icon to be visible when the width of a column is less than the width of the text content of the header.

### Steps to reproduce  

1. Open the [Custom Column Pinning UI sample](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-pinning#custom-column-pinning-ui) in StackBlitz
2. Make the width of a column to be less than the width of the header's text content.

### Result  
The pin icon will be cut off.

### Expected result  
The pin icon should be visible.

### Attachments  
image

viktorkombov:
Closes #3163, #3182

### DIFF
--- a/live-editing/configs/HierarchicalGridConfigGenerator.ts
+++ b/live-editing/configs/HierarchicalGridConfigGenerator.ts
@@ -342,7 +342,7 @@ export class HierarchicalGridConfigGenerator implements IConfigGenerator {
         }));
 
         configs.push(new Config({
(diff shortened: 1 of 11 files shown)


dkamburov:
no need to pr directly to master

PR #3191, opened 2022-12-05
Create dropdown hierarchical selection samples

PlamenaMiteva:


src/app/data-entries/dropdown/dropdown-tree-hierarchical-selection/dropdown-tree-hierarchical-selection.component.ts
```diff
@@ -0,0 +1,35 @@
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { IBaseChipEventArgs, IgxTreeComponent, ITreeNodeSelectionEvent } from 'igniteui-angular';
+import { SINGERS } from '../../../data/singersData';
```
kdinev, line 3: For this sample, use different kind of data. The data is not homogeneous and that's why there was custom logic to not take text from every selected node. Instead, the data should be homogeneous and the selection shouldn't cascade, so it better displays a real-world scenario :)

kdinev (changes requested):

src/app/data-entries/dropdown/dropdown-tree-hierarchical-selection/dropdown-tree-hierarchical-selection.html
```diff
@@ -0,0 +1,20 @@
...
+      <igx-chip *ngFor="let node of selectedNodes" [id]="node.data.ID" [removable]="true" (remove)="chipRemoved($event)">{{node.data.Name}}</igx-chip>
+    </igx-chips-area>
+
+    <button class="button" igxButton="raised" [igxToggleAction]="countriesDropDown" [igxDropDownItemNavigation]="countriesDropDown">
+      IgxTree
+    </button>
+</div>
+
+<igx-drop-down #countriesDropDown [height]="'500px'" [width]="'300px'">
+    <igx-tree #igxTree selection="Cascading" (nodeSelection)="onNodeSelection($event)">
```
kdinev, line 12: The tree should not be with cascading selection.

src/app/data-entries/dropdown/dropdown-tree-hierarchical-selection/dropdown-tree-hierarchical-selection.html
```diff
@@ -0,0 +1,20 @@
+<div class="button-wrapper">
+    <igx-chips-area>
```
kdinev, line 2: Wrap the chip area and button in a container which leaves empty space for the chip area when there's no chips in it. Also replace the raised button with an icon button with chevron down or something of that sort.

kdinev (changes requested):

src/app/data-entries/dropdown/dropdown-tree-hierarchical-selection/dropdown-tree-hierarchical-selection.html
```diff
@@ -0,0 +1,21 @@
+<div class="container">
+    <div class="chip-container">
+        <igx-chips-area>
+            <igx-chip *ngFor="let node of selectedNodes" [id]="node.data.ID" [removable]="true" (remove)="chipRemoved($event)">{{node.data.Name}}</igx-chip>
+        </igx-chips-area>
+    </div>
+    <button igxButton="icon" [igxToggleAction]="countriesDropDown" [igxDropDownItemNavigation]="countriesDropDown">
+        <igx-icon fontSet="material">expand_more</igx-icon>
+      </button>
```
kdinev, line 9:
```suggestion
    <button igxButton="icon" [igxToggleAction]="countriesDropDown" [igxDropDownItemNavigation]="countriesDropDown">
        <igx-icon fontSet="material">expand_more</igx-icon>
    </button>
```

kdinev (changes requested):

jackofdiamond5 (commented):
We should also make sure that the view file names, and the style file names have a `component` in them, e.g. `dropdown-tree-hierarchical-selection.scss` => `dropdown-tree-hierarchical-selection.component.scss`

jackofdiamond5:
I'm thinking - right now every time we interact with anything other than the drop-down, it will immediately close. This includes when clicking on a chip and attempting to remove it. 
Like so:
close-dd

This, however, does not happen if you use keyboard navigation to go to the chip and deselect it with `Space` / `Enter`:
not-close-dd

So, I'm thinking, maybe we can leave the dropdown open as long as the user is interacting specifically with the chips no matter if they're using the keyboard or the mouse, like this:
not-close-dd1

This should be somewhat easy to implement, like for example we could look at the drop-down's closing event target's composed path and then cancel the event if there's an `igx-chip` in the path. This should allow us to keep the drop-down open if we click anywhere on the chip, while at the same time have the drop-down close itself if we click anywhere else. @kdinev what do you think?

src/app/data-entries/dropdown/dropdown-tree-grid-hierarchical-selection/dropdown-tree-grid-hierarchical-selection.component.ts
```diff
@@ -0,0 +1,42 @@
...
+                this.igxTreeGrid.deselectRows([row.ID]);
+            }
+            return row.ID !== event.owner.id;
+        });
+    }
+
+    public handleClosing(event: IBaseCancelableBrowserEventArgs) {
+        var eventTarget = event.event.target as HTMLElement;
+        if (eventTarget.parentElement.classList.contains('igx-chip__remove')) {
+            event.cancel = true;
```
jackofdiamond5, line 39: You can simplify this to `evt.cancel = evt.event.composedPath().some(e => (e as HTMLElement).nodeName.toLowerCase() === 'igx-chip');`. I think getting the composed path is preferable because it will tell us all of the elements the event has gone through up until now. This will allow for chips to be clicked on as well, not just their remove icon. The same should go for the other sample too.

jackofdiamond5 (approved):

Lipata:
Merged in nNext branch and will get to production with a mass merge.

PR #3198, opened 2022-12-07
feat(input-group): add validation samples

### LINKED ISSUE #3192: Add input-group validation samples
Related to https://github.com/IgniteUI/igniteui-docfx/issues/3146

RivaIvanova:
Closes #3192


src/app/data-entries/input-group/reactive-form-custom-validation/reactive-form-custom-validation.component.ts
```diff
@@ -0,0 +1,32 @@
...
+        username: new FormControl<string>('', { nonNullable: true, validators: [this.usernameValidator('infragistics')] })
+    });
+
+    get username() {
+        return this.registrationForm.get('username');
+    }
+
+    private usernameValidator(val: string): ValidatorFn {
+        return (control: AbstractControl): ValidationErrors | null => {
+            if (control.value?.toLowerCase() === val) {
```
kdinev, line 25: Make this a little more robust and check if the username contains the value as an email extension, for example.

src/app/data-entries/input-group/reactive-form-custom-validation/reactive-form-custom-validation.component.html
```diff
@@ -0,0 +1,9 @@
+<article class="sample-column">
+    <form [formGroup]="registrationForm">
+        <igx-input-group>
```
kdinev, line 3: Extend this sample to one more input. There should be a password and repeat password field. The password should validate based on `minLength`, existence of numeric and special characters `pattern`, while repeat password should validate based on being the same as the password field. 

UPDATE: actually, it would be cool to check that the password doesn't contain the username and if it does, flag that it shouldn't. This would be a custom validator, not just pattern and minLength.

kdinev (changes requested):

src/app/data-entries/input-group/reactive-form-custom-validation/reactive-form-custom-validation.component.ts
```diff
@@ -59,75 +60,66 @@ export class ReactiveFormCustomValidationComponent {
...
-    private usernameValidator(val: string): ValidatorFn {
+    private emailValidator(val: string): ValidatorFn {
         return (control: AbstractControl): ValidationErrors | null => {
-            if (control.value?.toLowerCase().includes(val)) {
-                return { username: true };
+            const value = control.value?.toLowerCase();
+            const localPartRegex = new RegExp(`(?<=(${val})).*[@]`);
+            const domainRegex = new RegExp(`(?<=[@])(?=.*(${val}))`);
+
+            const returnObj = {};
```
kdinev, line 69: Make an interface describing the error object and assign that type to the return object. Then you wouldn't have to use `['prop']` notation and it also looks cleaner!

kdinev (changes requested):

jackofdiamond5 (approved):

PR #3252, opened 2023-02-21
Promote Staging to Production

hanastasov:


ChronosSF:
These seems to be merging master into vNext

hanastasov:
> This seems to be merging master into vNext

Good catch :D

zdrawku:
@hanastasov @ChronosSF btw guys we spoke with @damyanpetev that the mass merges should be applicable both ways. For example, Damyan did a change in master and his expectation is to be able to mass merge it into vNext anytime.

PR #3258, opened 2023-02-23
fix(hierarchical grid): add richer data and a few new columns

### LINKED ISSUE #3250: Set all Grid samples width to be 100%
### Description  
Previously we had to set a fixed width in order for the Grid to fit the iframe, but now the iframe is sized based on the codeview container and its okay to set grid's height and its parent element to be 100% in order to take the full size within the code view and at the same time be with full screen size when opened alone. 

Example with CRM Grid:
image

Full screen:
image

monacodelisa:
Closes #3250

### DIFF
--- a/src/app/data/hierarchical-data.ts
+++ b/src/app/data/hierarchical-data.ts
@@ -395,19 +395,28 @@ export const CUSTOMERS = [{
             ProductID: 11,
             UnitPrice: 14.0000,
             Quantity: 12,
-            Discount: 0.0000000e+000
+            Discount: 0.0000000e+000,
+            Weight: 2.5,
+            Length: 25.6,
+            TotalPrice: 225.75
         },
         {
             ProductID: 42,
             UnitPrice: 9.8000,
             Quantity: 10,
-            Discount: 0.0000000e+000
+            Discount: 0.0000000e+000,
+            Weight: 2.1,
+            Length: 29.3,
+            TotalPrice: 250.50
         },
         {
             ProductID: 72,
             UnitPrice: 34.8000,
             Quantity: 5,
-            Discount: 0.0000000e+000
+            Discount: 0.0000000e+000,
+            Weight: 2.7,
+            Length: 24.8,
+            TotalPrice: 220.25
         }
         ]
     }]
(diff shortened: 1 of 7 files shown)


src/app/hierarchical-grid/hierarchical-grid-column-pinning/hierarchical-grid-toolbar-pinning.component.ts
```diff
@@ -9,15 +9,16 @@ import { ActivatedRoute } from '@angular/router';
 })
 export class HGridToolbarPinningComponent implements OnInit {
     public localdata;
-    public dark = false;
+    public dark = "";
```
hanastasov, line 12: I see why it is done this way (setting it as string) and it works out nicely when you receive a string value from the query params.

In general, IMO, it is good to keep the value boolean, and the check whether it is a "true" string being moved from the template to the place where you set the value on line 21:

```
this.dark = params.dark === 'true' ? true : false;
```

src/app/data/hierarchical-data.ts
```diff
@@ -519,17 +543,26 @@ export const CUSTOMERS = [{
...
+            Discount: 5.0000001e-002,
+            Weight: 3.1,
+            Length: 24.3,
+            TotalPrice: 219.25
         }, {
             ProductID: 65,
             UnitPrice: 16.8000,
             Quantity: 20,
-            Discount: 0.0000000e+000
+            Discount: 0.0000000e+000,
```
hanastasov, line 562: Why setting the discount in this format, seems curious

src/app/grid/grid-sample-pinning/grid-toolbar-pinning.component.html
```diff
@@ -1,4 +1,4 @@
-<div [ngClass]="{'grid__wrapper': true, 'dark-theme': dark === true }" style='width: 100%;'>
+<div [ngClass]="{'grid__wrapper': true, 'dark-theme': dark === 'true' }" style='width: 100%;'>
```
hanastasov, line 1: After keeping the `dark` value boolean, you will be able to use a simpler expression here:

```
<div [ngClass]="{'grid__wrapper': true, 'dark-theme': dark }" style='width: 100%;'>
```
Then it seems even nicer to rename the constant to smth more descrptive, for example `useDarkTheme`, and then the template seems even more readable:

```
<div [ngClass]="{'grid__wrapper': true, 'dark-theme': useDarkTheme }" style='width: 100%;'>
```

src/app/hierarchical-grid/hierarchical-grid-column-pinning/hierarchical-grid-toolbar-pinning.component.ts
hanastasov: > 

Null coalescing operator serves a more specific purpose, so when I think about it the Logical OR operator would serve us best here.

src/app/data/hierarchical-data.ts
monacodelisa: This is not a change applied by me I only added the properties below it

src/app/grid/grid-sample-pinning/grid-toolbar-pinning.component.html
monacodelisa: Ok I'm on it

hanastasov (approved):

Issue #3295, opened 2023-04-12
SignalR sample has UI that is not readable in dark mode


kdinev:
### Description  
When I open the grid finjs sample, the default mode is dark. Market data values that are positive are not readable in it. Also the combos to switch options are enormous!

### Steps to reproduce  

1. Open https://www.infragistics.com/angular-demos-lob/grid-finjs-dock-manager/grid-finjs-dock-manager
2. Observe the top-right grid.

### Result  
Positive results in the `Price`, `Change` and `ChangeP` columns are not readable.

### Expected result  
The UI should be readable in both dark and light mode.

### Attachments  
![image](https://user-images.githubusercontent.com/1472513/231465669-9b0674fc-1808-47c5-bb56-644997e5c529.png)

PR #3326, opened 2023-05-18
fix(dock-manager): change outlet's class when the theme is toggled

### LINKED ISSUE #3296: SignalR sample doesn't switch paging drop-down to light mode
### Description  
When I switch to light mode in the grid finjs sample, the paging drop down stays dark mode.

### Steps to reproduce  

1. Open https://www.infragistics.com/angular-demos-lob/grid-finjs-dock-manager/grid-finjs-dock-manager
2. Switch to light mode.
3. Change bottom-right grid page size.

### Result  
The page size drop down is still in dark mode.

### Expected result  
The page size drop down also switched to light mode.

### Attachments  
image

viktorkombov:
Closes #3296, #3295


hanastasov (approved):

hanastasov (changes requested):
image

projects/app-lob/src/app/grid-finjs-dock-manager/grid-finjs-dock-manager.component.scss
```diff
@@ -127,6 +127,12 @@
 	}
 }
 
+.dark-theme {
```
hanastasov, line 130: Hm, maybe can be fixed in the dark-theme itself, rather than per this sample?

hanastasov (changes requested):

hanastasov (approved):

PR #3327, opened 2023-05-22
Updating PROD with latest fix and enhacements

hanastasov:
Resolving:

**FinJS Dock manager sample:**
- add the right selector for the selected columns styling
- change styles for the disabled chip****
- make the UI  (for selected column and positive values) readable in dark mode
- change the background color of the selected columns in grids
- change display density of select components to compact
- change outlet's class when theme is toggled

**Other samples:**
- update carousel sample alignment
- fix sample configs (


dobromirts (dismissed):

projects/app-lob/src/styles.scss
```diff
@@ -29,8 +29,15 @@ body {
     .grid-chart-contextmenu-wrapper {
         @include fluent-dark-theme($fluent-excel-palette);
     }
+    igx-grid,
```
hanastasov, line 32:
```suggestion
    // remove this workaround after resolving https://github.com/IgniteUI/igniteui-angular/issues/13039
    igx-grid,
```

PR #3402, opened 2023-11-08
chore(*): update 17

Lipata:


zdrawku:
@Lipata, @dobromirts is aware of the problem with the live-editing build, he will submit a fix shortly and re-test

zdrawku:
Actually, the problem comes from the library itself, as @ChronosSF pointed out "We should either add a migration for this, add it in the migration guide or make the parameter optional, if possible"

kdinev:
@zdrawku I will update the live-editing.

zdrawku:
We can start testing on staging once the PR is merged

ChronosSF (approved):

PR #3430, opened 2024-02-07
update 17.1.x

Lipata:
Update ng@17.2.x, igniteui-angular@17.1, igniteui-angular-charts@17.2


Lipata:
@zdrawku @dobromirts I cannot install 'igniteui-angular-extras' without --force, when trying to update with the next version for [igniteui-angular](https://github.com/IgniteUI/igniteui-angular/releases/tag/17.1.0-beta.0), which is 17.1 compatible. The same is valid if I try to update to latest [igniteui-angular-charts@17.2.x](https://www.npmjs.com/search?q=igniteui-angular-charts). I guess we need new beta for igniteui-angular-extras with both official DV packages and latest next version for igniteui-angular.

kdinev (approved):

Lipata:
@SisIvanova with the latest commit the `igniteui-angular` was updated and the button migrations were applied. I only see one problem related to the `igxButtonBackground` property:

```console
error NG8002: Can't bind to 'igxButtonBackground' since it isn't a known property of 'button'.
```
I see that it is supposed only to be deprecated in 17.1? Or we will remove it and add a migration?

PR #3456, opened 2024-04-26
feat(forms): updating the stepper sample to use reactive forms

Marina-L-Stoyanova:


valadzhov:
In the third step if the checkboxes are checked and then unchecked the button "Contine" remains enabled and it is possible to continue to the next step, this behaviour is rather strange:
image

PR #3486, opened 2024-06-04
fix(igxGrid): Temporary fix flickering by setting an explicit row hei…

MayaKirova:
…ght since it does not get calculated correctly in remote data scenario when --ig-size  is changed.


MayaKirova:
Closing since there's a fix: https://github.com/IgniteUI/igniteui-angular/pull/14331

PR #3488, opened 2024-06-06
Update to rc.1

ChronosSF:


src/app/lists/combo/combo-remote/combo-remote.component.html
```diff
@@ -10,6 +10,6 @@
     (closing)="onClosing()" (opened)="onOpened()"
     (closed)="onClosed()"
     placeholder="Location(s)"
-    [filterable]="true">
+    [filteringOptions]="{ filterable: true, caseSensitive: false }">
```
damyanpetev, line 13: These are just the default values, can be omitted entirely, same for the styling sample //CC @Lipata

PR #3571, opened 2024-11-26
Ttonev/hr portal sample

tishko0:


yradoeva:
Here are the things I noticed, related to the design:

1. Can we make the so-called “zebra” styling of the rows by making the row-even-background color surface light (ig-surface-500)?
Executive

2. The gap between flags and countries/cities and between the buttons in Contacts column should be 8px (not 10px)
tria

3. There should be a gap of 8px also between the Avatar and the names in the Name column
John Smith

4. “Job Title” and “Gross Salary” should have an interval

5. The flags should be smaller - 18x14px (not 24x18px)
6. The flags should have rounded corners with border-radius of 1px and a shadow - Elevations/Shadow.01
Pasted Graphic 9

7. Hire Date - for some reason all months appear as “00”, but they shouldn’t
Hire Date

8. The Quick Filter is missing:
Pasted Graphic 5

9. Gross Salary header cell should be right-aligned
Gross Salary

**Theming Related:**
1. Font should be Open Sans for Fluent theme
2. All Buttons are with All Caps, but they shouldn’t. That applies to the buttons in the Toolbar, in Excel Style filtering and in Advanced filtering dialogs
_By design:_ 
image

PR #3600, opened 2025-02-24
Argument palette removed from color-classes mixin

ivanvpetrov:
Argument **palette** is no longer used and throws errors on build


simeonoff:
@ivanvpetrov migrations have been added to 19.1.x for this issue. All you have to do is run them.

PR #3631, opened 2025-03-11
feat(query-builder): add SQL query sample

igdmdimitrov:
Related to [#6054](https://github.com/IgniteUI/igniteui-docfx/issues/6054)


src/app/interactions/query-builder/query-builder-sql-sample/query-builder-sql-sample.component.html
```diff
@@ -0,0 +1,16 @@
+<div class="wrapper">
+    <div class="container">
+        <igx-query-builder #queryBuilder
```
gedinakova, line 3: How about adding the `output-area` class to QB:
```suggestion
        <igx-query-builder #queryBuilder class="ouput-area"
```
This way we get the same border as the one around the SQL query:

image

gedinakova (approved):

gedinakova (approved):

ChronosSF (dismissed):

PR #3648, opened 2025-03-19
QB sample requesting data from an endpoint

gedinakova:


ChronosSF:
@gedinakova , @teodosiah , when you start switching columns from the Select dropdown on and off the grid doesn't recreate/remove them properly.

ChronosSF (approved):

PR #3725, opened 2025-08-28
feat(grids): added both side pinning samples

tishko0:
20.1 feature, in order to work please use "igniteui-angular": "20.1.0-alpha.1" or later version


ChronosSF (changes requested):
Please add live-editing configuration for these samples.

ChronosSF (dismissed):

PR #3761, opened 2025-10-27
Ttonev/fix both side pin sample

### LINKED ISSUE #3758: Pinning on Both sides sample  issues and improvements
The sample needs some improvements and has an issue:

- open prod version
- select a column
- pin it to the left
- pin it to the right

The pinning border on the left disappears, this happens in React too.

Also if the buttons are not supposed to work for specific states of the column or grid, then bind their disabled property to something that checks the state. E.g. if a column pinned to the right cannot be directly pinned to the left with the button but instead has to be unpinned first, then you can disable the pin right button when selecting on it. Same for a pin left button if you select a column already pinned to the left.

tishko0:
fixes #3758


src/app/hierarchical-grid/hierarchical-grid-sample-both-sides-pinning/hierarchical-grid-both-sides-pinning.component.ts
```diff
@@ -71,13 +71,15 @@ export class HierarchicalGridBothSidePinningSampleComponent implements OnInit {
 
     public pinLeft() {
         this.grid1.selectedColumns().forEach((col: IgxColumnComponent) => {
+            col.pinned = false;
```
MayaKirova: Also here pls

PR #3895, opened 2026-02-12
feat(PDF): New sample for using a custom font

gedinakova:


mtsvyatkova:
The file upload option is not clearly distinguishable. To upload a font, you need to click on “Upload a custom font (.ttf)” (the Choose File field).

image

PR #3896, opened 2026-02-17
refactor(samples): replace css-vars with tokens mixin

didimmova:


simeonoff:
This PR is not needed.

PR #3913, opened 2026-02-23
fix(LOB): fix minor issues in LOB

### LINKED ISSUE #3912: Minor issue in LOB samples
Start the LOB sample:
- navigate to `Pivot Grid With Datasource Component`. There is no pivot grid on the view only the pivot selector is rendered.
- navigate to `Grid Finance partial demo sample`. Both `Grid Finance demo sample` and `Grid Finance partial demo sample` are marked as active paths.

wnvko:
Fix finjs grid navigation paths and flex container showing pivot grid.
Closes #3912

### DIFF
--- a/src/app/pivot-grid/pivot-datasource/pivot-data-selector-sample.component.html
+++ b/src/app/pivot-grid/pivot-datasource/pivot-data-selector-sample.component.html
@@ -1,11 +1,8 @@
 <div class="pivot-container">
-    <div>
+    <div style="flex-grow: 1;">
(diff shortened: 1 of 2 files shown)


src/app/pivot-grid/pivot-datasource/pivot-data-selector-sample.component.html
```diff
@@ -1,11 +1,8 @@
 <div class="pivot-container">
-    <div>
+    <div style="flex-grow: 1;">
```
gedinakova: Not sure we even need the inner div, but not something to really bother about.

ChronosSF (approved):

PR #3922, opened 2026-03-09
Fix disable summaries sample logic

### LINKED ISSUE #3892: [Grid] Disable Summaries Design is not intuitive
1. Open this sample https://staging.infragistics.com/angular-demos/grid/grid-disable-summaries
2. The UX seems a bit confusing. When you select an item from the dropdown it actually hides the selected summary. It would be much more intuitive to SHOW the selected items and hide the non-selected.
3. I understand the issue is about disabled summaries, but we don't need to have everything too literally translated into the code. Think about it, when you go on some website and click on some checkbox, do yo expect it to show or to hide content related to it?

Image

MarielaTihova:
Resolves #3892

### DIFF
--- a/src/app/grid/grid-disable-summaries/grid-disable-summaries.component.html
+++ b/src/app/grid/grid-disable-summaries/grid-disable-summaries.component.html
@@ -1,6 +1,6 @@
 <div class="grid-wrapper">
   <div class="summaries">
-    <h5 class="summaries-title">Disable Summaries for Column:</h5>
+    <h5 class="summaries-title">Toggle Summaries for Column:</h5>
     @for (column of columns; track column; let i = $index) {
       <div class="summary-column-button">
         <button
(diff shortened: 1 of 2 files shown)


georgianastasov (changes requested):
For now I reject this pull request and it should not be merged without the opinion of the UI/UX person, more precisely without the opinion of @sdimchevski and @yradoeva because they are the people who created the sample design. We have a lot of discussions about that when the feature and sample were implemented and the final decision was that behavior: "_all checkboxes enabled at the beginning and checking disables the summary, this is what most corresponds to the name of the property and what we want to show_".

MarielaTihova:
> For now I reject this pull request and it should not be merged without the opinion of the UI/UX person, more precisely without the opinion of @sdimchevski and @yradoeva because they are the people who created the sample design. We have a lot of discussions about that when the feature and sample were implemented and the final decision was that behavior: "all checkboxes enabled at the beginning and checking disables the summary, this is what most corresponds to the name of the property and what we want to show".

In my opinion the name should be changed. It is in no way intuitive to have checkboxes checked when something is disabled and unchecked checkboxes when something is enabled. It was confusing for me as a developer. I imagine it would be even more confusing for the users. I accept feedback from @sdimchevski  and @yradoeva of course.

sdimchevski:
I don’t fully agree that the current UX is confusing. The dropdown is labeled "Disabled Summaries" which clearly indicates that the user is selecting summaries to hide not summaries to show. Because of that the behavior is consistent with the label: when a summary is selected it gets disabled and therefore disappears from the footer.

This approach can also be more efficient since all summaries are visible by default and the user only needs to disable the few they don’t need rather than manually selecting every summary they want to display.

dkamburov:
In that case we can close the pr and the issue. Thanks!

Issue #3936, opened 2026-04-21
Advanced filtering custom styling sample does not apply custom styles.


MayaKirova:
Custom styling sample at:
https://staging.infragistics.com/products/ignite-ui-angular/angular/components/grid/advanced-filtering#demo-1

Shows no custom styling anymore. It shows the default styles.

The same sample on live shows custom styles correctly: 
https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/advanced-filtering#demo-1

PR #3962, opened 2026-05-15
fix(grid): fix advanced filtering grid styles

Hristo313:
Revert the old custom styles for the advanced filtering

Closes https://github.com/IgniteUI/igniteui-angular-samples/issues/3936


dkamburov:
@simeonoff Is it ok to revert to the old custom styles for that sample? Or we should try different approach?

@Hristo313 hgrid and treegrid adv. filtering samples also needs styling fixes.

simeonoff:
No it's not ok, in fact I will close this PR now in favor of https://github.com/IgniteUI/igniteui-angular-samples/pull/3981. The PR shows how to alter the styles of the advanced filtering menu using a different approach that doesn't require as many themes or tokens to be set.

PR #3963, opened 2026-05-19
fix(theming): fix theming bootstrap dropdown

### LINKED ISSUE #3810: Bootstrap theming sample issues
Open the theming sample at:
https://staging.infragistics.com/products/ignite-ui-angular/angular/components/themes/misc/bootstrap-theming#demo

- Expanding the Accordion items below show data, but then hides it and shows as empty:

Image

- Also the topic and screenshots include a theming button, but I don't see one in the actual sample.

Hristo313:
Closes #3810 

The conflict seems to come from ngbAccordion/ngbAccordionCollapse interaction with overlay-based components (igx-dialog, dropdowns).
Replacing the accordion with a lightweight local toggle implementation removes the DOM manipulation/collapse behavior that was interfering with dropdown rendering and interaction

### DIFF
--- a/src/app/theming/bootstrap/bootstrap-sample.component.html
+++ b/src/app/theming/bootstrap/bootstrap-sample.component.html
@@ -227,13 +227,11 @@
                   </div>
                 </div>
                 <div class="sample-column">
-                  <div ngbAccordion>
-                    <div ngbAccordionItem>
-                      <h2 ngbAccordionHeader class="m-0">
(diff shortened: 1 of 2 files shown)


skrustev (changes requested):
Removing the `ngbAccordion` directives and using custom implementation directly doesn't sound like a fix to me. Either this should be logged as a product bug or if there's a way to make it work in conjunction with the Bootstrap Theming, which is the point of the sample anyway.

PR #4015, opened 2026-08-18
Allows for the theming widget to style samples

ChronosSF:


ChronosSF:
@copilot+gpt-6-astra , the input - related changes are no longer okay, restore them so that the PR only includes the theme picker -related code.

PR #4020, opened 2026-08-25
refactor(button-group): update styles and simplify layout structure across all samples

desig9stein:


Hristo313:
In the States sample the disabled buttons are not disabled:
image

In the Alignment sample the labels should be centered:
image

Hristo313 (dismissed):

ChronosSF (approved):

